PR #4, opened 2020-07-29
Escape square brackets in regex

lundwall:
Square brackets in a regular expression match any single character enclosed, and here we want all of them to match correctly.


TRaSH-:
@lundwall  thnx

PR #5, opened 2020-08-05
Update How-to-get-Quality-Releases.md - Custom Formats

bakerboy448:
Rough instructions on creating a custom format


TRaSH-:
i'll deny this upgrade and not because it's borked
but because you got in your instructions `To create a custom format in Radarr V3.....`
and these part of the guide is for Radarr v0.2
Adding Radarr V3 today,
if you find something then i'll be happy to add it.
being Radarr V3 Custom Formats is a bit more complicated so any extra description/instruction would be very helpful.

nemchik:
V3 repo being merged into this one?

nemchik:
NVM, I see it's already done. Nice!

PR #6, opened 2020-08-09
Update Collection-of-Custom-Formats-for-Radarr.md

owine:
Add link to Single Layer DV, revise Single Layer DV format (revised by @TRaSH-)


TRaSH-:
I Had my email client setup wrong so didn't get a notification for this,
sorry for this fail on my side.
I think you messaged me in Discord and it should be already added to the Guide and also updated it.
Let me know if it's okay now else you can open a new issue.
Thnx

PR #10, opened 2020-09-06
Update Collection-of-Custom-Formats-for-RadarrV3.md

bakerboy448:
add -YTS to Block 1 (YIFY AKA)

Should these be "YTS" and "YIFY" the releases are not always "-YTS" or "-YIFY"?


TRaSH-:
I was just looking through them.
Will add it tomorrow.
Couldn't merge them.
Might even remove a few `-` being that CF handles it better

bakerboy448:
Also some various typos and inconsistencies to fix; If you don't catch them, I'll edit after tomorrow's update

PR #17, opened 2020-09-11
Escape periods

nemchik:
There's a lot of places where you have `.?`

`.` means any character
`?` means 0 or 1

So the result is "0 or 1 or any character" but based on the context I believe you want `\.?` which is "0 or 1 of a literal period character"

In some other places you already have this done, but these are the ones I found that did not.


TRaSH-:
Can't add this being I had to redo the complete guide

PR #54, opened 2020-12-12
Update Sonarr-Release-Profile-RegEx.md

TRaSH-:
Updated: Sonarr Release Profile RegEx (WEB-DL)
- Fixed: block x265 for 1080p being it didn't recornize when their was a period between the x265
- Cleanup: Hidden the Changes Updates by defaults, added some extra info to try to make it more clear


kdbruin (commented):
The new regex will allow any character between x265 or h265. Better to use `\.?` instead.

TRaSH-:
Thnx
I'll take a look at it when I get home.
Weird I didn't notice any of this with the 2-3 tests I've done

PR #59, opened 2020-12-13
Update how-to-set-up-hardlinks-and-atomic-moves.md

TRaSH-:
Updated: How To Set Up Hardlinks and Atomic-Moves
- Updated: the Synology section with:
  - Added permission settings
  - Added cd to /volume1
- Updated: the Docker section with:
  - changed order when to set permissions settings


TRaSH-:
double made

PR #98, opened 2021-01-17
Create FUNDING.yml

TRaSH-:


TRaSH-:
Seems i messed up somewhere double entry

PR #151, opened 2021-02-19
Radarr-CF-Collection-Split-Script

TRaSH-:
Splitted up Custom Format Collection to make it workable with the auto add script

@rcdailey

### DIFF
--- a/docs/Radarr/V3/Radarr-CF-Audio-Advanced.md
+++ b/docs/Radarr/V3/Radarr-CF-Audio-Advanced.md
@@ -1,0 +1,1294 @@
+# Radarr-CF-Audio-Advanced
+
+------
+
+## Audio Advanced
+
+------
+
+### TrueHD ATMOS
+
+??? example "json"
+
+    ```json
+    {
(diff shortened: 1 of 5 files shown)


rcdailey (commented):
I was expecting removals in the original custom formats markdown file. Am I missing something?

TRaSH-:
As we talked in discord about it, I won't be doing that anytime soon.
Separate them will make it less clear in my opinion and a pain to direct users to it when providing support, and it would/could break other guides i created that have redirects to the main page to a certain Custom Format.
Including links we got in Radarr support to help people out. (For example look at page x or was it y)

I created these just plain for your script for the time being (they don't show up in the guide as separate),
I might consider to switch to separate pages, But only if it doesn't break the readability and usability of the main guide being that's the main factor of the guide in the first place !

rcdailey:
I think based on our conversation we can go ahead and close this one. I think instead we will focus on providing scores in the text above each CF, should there be a preset score you recommend for them.

Thank you.

PR #152, opened 2021-02-23
Rename remux qualities to match Radarr

rcdailey:
Certain remux quality names did not exactly match the names used in
Radarr. Modified names are:

* `Bluray-1080p Remux` -> `Remux-1080p`
* `Bluray-2160p Remux` -> `Remux-2160p`

Markdown table formatted for styling as needed after the above changes.


TRaSH-:
Seems I just plain copy/pasted it from Sonarr, that's why it didn't match Radarr.

PR #172, opened 2021-03-08
Add nhanc3 to list of low quality groups

jasonla:
PR to add nhanc3. They seem to re-encode with lower quality.


TRaSH-:
I did a search and found the same results you mentioned, but i also noticed that the file size are really low, and i think when you use the size quality from the guide you shouldn't get them in the first place.
but i will merge it.

PR #188, opened 2021-03-31
Update Basic DTS / Not DTS-HD regex

jwoodhouse:
Updates the Not DTS-HD regex to match against DTS-HD HRA, stops releases with DTS-HD HRA from being matched against both DTS and DTS-HD HRA custom formats.


TRaSH-:
I'll review it when I get home,
And check if other custom formats needs also some fixes like the ones we had with DD+

PR #189, opened 2021-04-01
Optional suggestions for preferring Dub releases when no dual-audio available 

g4m3r7ag:
Apologies if I've done something incorrect here, first time I've done anything other then open an Issue on GitHub. 

I noticed I was still getting Erai-raws releases when a dub was available but no dual audio option was. I had followed the initial instruction to not add the last two lines to the Must Not Contain but they were still being outscored due to those lines not being added anywhere in preferred. This was my solution. Dub releases appear to be getting a score of 406 now if they match both of the changes I made plus the +5 score in your optional section, so they're not outscoring a dual audio release with 501 or any of the ones with higher scores, which I assume are likely to be dual-audio if available. Not sure if the scoring should be modified.

### DIFF
--- a/docs/Sonarr/V3/Sonarr-Release-Profile-RegEx-Anime.md
+++ b/docs/Sonarr/V3/Sonarr-Release-Profile-RegEx-Anime.md
@@ -207,7 +207,7 @@ The reason most of these are added is due to their shitty quality or just in gen
 
 !!! important
 
-    If you would like `dub` releases then remove the last 2 lines and do not add them.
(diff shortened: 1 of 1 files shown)


TRaSH-:
@Visorask could you take a look at it and give your opinion about it.

Visorask:
Yeah I'll take a look tomorrow gonna be adding specific profiles for subs/dual-audio/dub, just doing a bunch of research first, also gonna be adding a guide for uncensored/uncut anime, as there are only so many releases and groups that do them.

TRaSH-:
Closed because of no updates after 3 months,
re-open when needed.

PR #208, opened 2021-04-18
Anime RP: Organize optional into its own section

rcdailey:
This makes parsing out optional terms much easier. "Optional" is in the
name of the header itself, which means anything within that section will
be excluded by default by TrashUpdater.

I've also organized it into its own release profile for clearer
separation. Should make it easier for users to update the profile
selectively since they don't need to dig through the numerous other
items in the Second Release Profile.

Other minor changes:

- Removed usage of admonition for many of the sentences since the TrashUpdater script skips these (they are meant to be informational to users, but not contain important data to Sonarr). This is an attempt to keep things in sync with WEB-DL.

### DIFF
--- a/docs/Sonarr/V3/Sonarr-Release-Profile-RegEx-Anime.md
+++ b/docs/Sonarr/V3/Sonarr-Release-Profile-RegEx-Anime.md
@@ -250,47 +250,6 @@ Add this to your Preferred with a score of **[501]**
 
 ---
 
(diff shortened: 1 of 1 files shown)


rcdailey:
@Visorask Feel free to steal this and update it to your liking if you don't like what I've done here.

rcdailey:
Closing in favor of @Visorask's PR #209

PR #231, opened 2021-05-22
Added to note.

sadstan123456:
Clarifies meaning of 'as high as possible'


sadstan123456:
There appears to be an issue with lines 61,62

sadstan123456:
Issue has been resolved

PR #233, opened 2021-05-22
Minor edit

sadstan123456:
Corrected capitalization


TRaSH-:
I actually always try to keep the casing as it's in the GUI

image

sadstan123456:
I seem to have missed that. It is good to follow the GUI's capitalisation.

PR #259, opened 2021-06-08
Update extra.css

nemchik:


nemchik:
Closing this for now. I probably have the right attribute, but the wrong value.

PR #262, opened 2021-06-10
Update README.md

doob187:
Add LSIO Plex Profile folder


TRaSH- (changes requested):
Being i hear different default paths for the LSIO container i suggest to add both paths like this.

`* LSIO container: `/appdata/plex/database/Library/Application Support/Plex Media Server/Profiles` or `/appdata/plex/Library/Application Support/Plex Media Server/Profiles`

PR #265, opened 2021-06-12
Update dependabot.yml

doob187:
create label dependabot or use own

image


TRaSH-:
closed for now, will reconsider if we actually need this

PR #273, opened 2021-06-18
Minor edit: Capitalisation

sadstan:


TRaSH-:
I always try to stay consistent how it's in the screenshots or on the original page,
If you change it like it's on the original wiki then i will approve and merge it.

image

sadstan:
Both earlier ones were wrong. Corrected now

PR #279, opened 2021-06-21
Add sonarr json and generator routine.

davidnewhall:
This contribution converts the markdown-esq Sonarr profiles into json files. A small python generator script is included to turn the JSON back into markdown. The code and idea was taken from @Roxedus  in [this commit](https://github.com/TRaSH-/Guides/compare/master...Roxedus:generator?diff=split). The "new" page looks almost identical to the previous. There are two specific ordering changes, explained in review comments.

To build the new templates just run `mkdocs build`. The templates are generated into `docs/Sonarr/generated` but they do not need to be committed. They are re-created every time `mkdocs` builds/deploys.

Please let me know if you have any questions!

- _Note1: Two of the new json files do not have a `trash_id`. Let me know what they should be and I'll add them._
- _Note2: The `Optional` section was not tested or setup properly because it's not used. We can fix that if needed._

PR #280, opened 2021-06-21
Add Sonarr JSON and generator routine (local branch).

davidnewhall:
Replaced #279. (see comments there)

This contribution converts the markdown-esq Sonarr profiles into json files. A small python generator script is included to turn the JSON back into markdown. The code and idea was taken from @Roxedus  in [this commit](https://github.com/TRaSH-/Guides/compare/master...Roxedus:generator?diff=split). The "new" page looks almost identical to the previous. There are two specific ordering changes, explained in review comments.

To build the new templates just run `mkdocs build`. The templates are generated into `docs/Sonarr/generated` but they do not need to be committed. They are re-created every time `mkdocs` builds/deploys.

Please let me know if you have any questions!

- _Note1: Two of the new json files do not have a `trash_id`. Let me know what they should be and I'll add them._
- _Note2: The `Optional` section was not tested or setup properly because it's not used. We can fix that if needed._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,8 @@ venv/
 mkdocs-dev-server.bat
 /docs/Notifiarr/preview.bat
 /docs/Notifiarr/Integrations/_TEMPLATE.md
+__pycache__
+/site
+generated
+.DS_*
+._*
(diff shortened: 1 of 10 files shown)


TRaSH-:
Closed,
looking for a better solution, or waiting till Sonarr finally will support decent import/export to json of their release profile

PR #283, opened 2021-06-28
Bazarr Base URL adjustment

Getn67:
Clean up of Base URL section. Emphasis on instances where a Base URL is needed, and where it isn't.


Getn67:
Adjusting. Missed a word.

PR #287, opened 2021-07-02
Minor edit

sadstan:
Re-ordered pastebins


TRaSH-:
i would add pastebin at the end, i'm actually considering to remove it,
being they starting to get worse and worse.
to many ads and they are also blocking pastes

sadstan:
I find pastes difficult to read on hastebin, hence I put it last.

TRaSH-:
> I find pastes difficult to read on hastebin, hence I put it last.

sure NP,
let's just remove pastebin, i just hate those ads. especially when i'm on my phone

Radarr uses 
0bin,
Use Hastebin,
Ubuntu's Pastebin,
Gist

So we could replace it for one of the others

sadstan:
Done.

PR #311, opened 2021-08-12
Update hq.json

KnifeFed:
NCmt listed twice


TRaSH-:
> NCmt listed twice

Let me review it tomorrow when I get home.
I already was doing some small updates

PR #312, opened 2021-08-14
Update Sonarr-recommended-naming-scheme.md

briandipalma:


TRaSH-:
Thnx for the Jellyfin update.

I made a few changes/suggestions and also have a few questions.

1. You changed optional to recommended. the reason why i used optional is because when in the worse case you get a sonarr database corruption/crash/loss and you need to re-import your existing media using the imdb in the series folder name could result in non matching results.
- I changed this back to optional.

2. A few things don't really match up being that jellyfin uses `imdbid-` where with plex we use `imdb-`
- I added a example where you see the results for Jellyfin.
- **Question:** Does Jellyfin also support `imdb-` that would make it easier and more consistent and for people easier to switch or use both ?

3. Removed the tmdb suggestion because `IMDb IDs are going to be very accurate and rarely change, TVDB/TMDB IDs, on the other hand, do change or are removed more frequently.` what would result in bad results.
and manually doing changes to folder names defeats the purpose of automation in my opinion.

4. Also did a few description changes/moves to make it less messy with the Jellyfin suggestion/example.

5. Do you know if the Jellyfin suggestion also work with Emby ?

Let me know what you think about the suggestions/changes, I'm open for any other suggestions/changes to make it more clear if needed. 

Example:

image

briandipalma:
Hi, thanks for responding.

I don't understand 1. Are you saying that Sonarr's DB could become corrupted and mismatch IMDB IDs e.g. instead of `10` matching `Lost` it matches `Breaking Bad`? I didn't realize that could happen, what a mess but surely that could happen even without the IDs in the names? Either way optional is fine, I just wanted to highlight that using IDs means Jellyfin/Plex etc don't have to guess what a piece of media is so it's generally a more robust naming scheme.

As for 2. yeah it's annoying that it's different but that's what I gather from the code: https://github.com/jellyfin/jellyfin/blob/f50bdb0f328e0c853f97521b12c3618b1af55e0b/Emby.Server.Implementations/Library/Resolvers/Movies/MovieResolver.cs#L362 I've raised a PR to update the Jellyfin documentation to confirm it https://github.com/jellyfin/jellyfin-docs/pull/559

The reason I mentioned 3. is due to a miniseries that was missing in TVDB and IMDB but was present in TMDB so I had to manually create the folder. I agree it's not ideal but at least now Jellyfin knows exactly what my media is. Specifically the 6 part Das Boot 1987 series which keeps getting deleted from TVDB as they confuse it with the 3 part 1985 recut of the movie. This shouldn't be a frequent occurrence so I'd suggest adding a warning around the suggestion saying it should be a last resort.

4. Cool

5. It appears to be almost the same as the Jellyfin style, which makes sense as it seems to have been there from before the fork: https://support.emby.media/support/solutions/articles/44001159102-movie-naming

TRaSH-:
> I don't understand 1. Are you saying that Sonarr's DB could become corrupted and mismatch IMDB IDs e.g. instead of `10` matching `Lost` it matches `Breaking Bad`? I didn't realize that could happen, what a mess but surely that could happen even without the IDs in the names? Either way optional is fine, I just wanted to highlight that using IDs means Jellyfin/Plex etc don't have to guess what a piece of media is so it's generally a more robust naming scheme.

No i'm not saying the database can get corrupted because you add the ID,
but the database could still get corrupted for any reason, like being stupid and use a network share for your database (not recommended) or a harddrive going bad, system crash during a write.

So if that happens and you need/want to re-import your collection in to Sonarr you actually want that it imports correctly without to many manual intervention.

> As for 2. yeah it's annoying that it's different but that's what I gather from the code: https://github.com/jellyfin/jellyfin/blob/f50bdb0f328e0c853f97521b12c3618b1af55e0b/Emby.Server.Implementations/Library/Resolvers/Movies/MovieResolver.cs#L362 I've raised a PR to update the Jellyfin documentation to confirm it [jellyfin/jellyfin-docs#559](https://github.com/jellyfin/jellyfin-docs/pull/559)

I also commented on that topic, let's hope they decide to support both options.

> The reason I mentioned 3. is due to a miniseries that was missing in TVDB and IMDB but was present in TMDB so I had to manually create the folder. I agree it's not ideal but at least now Jellyfin knows exactly what my media is. Specifically the 6 part Das Boot 1987 series which keeps getting deleted from TVDB as they confuse it with the 3 part 1985 recut of the movie. This shouldn't be a frequent occurrence so I'd suggest adding a warning around the suggestion saying it should be a last resort.

hmm, and how does sonarr handle this ?
being this is a Sonarr guide and not a Jellyfin specific one ?

I'll reconsider it if it needs to be added without running in to other problems.

> 5. It appears to be almost the same as the Jellyfin style, which makes sense as it seems to have been there from before the fork: https://support.emby.media/support/solutions/articles/44001159102-movie-naming

Okay so it might be best to change it to Jellyfin/Emby then

briandipalma:
> So if that happens and you need/want to re-import your collection in to Sonarr you actually want that it imports correctly without to many manual intervention.

I don't understand how adding a media DB ID would make it more likely to require an intervention...surely it would reduce the chance? It looks like Jellyfin/Plex and Emby all support these IDs in the folder names. This is done to remove ambiguity for them, why would it confuse Sonarr? Maybe the text could say it's recommended that you add a media DB ID with IMDB being the preferred one but that the format is media player specific as they all support different formats?

> I also commented on that topic, let's hope they decide to support both options.

The code seems straightforward enough, I think I could get it to support all the formats in Jellyfin.

> hmm, and how does sonarr handle this ?

It doesn't, sadly that series is outside Sonarr's control as it doesn't know it exists. It just ignores it, the ID is for the media players benefit. I guess it makes sense therefore to leave it out of this guide. As long as I can get it into the Jellyfin documentation I'm happy with that.

> Okay so it might be best to change it to Jellyfin/Emby then

We'll see if someone replies to my PR with an OK to the code changes.

TRaSH-:
> I don't understand how adding a media DB ID would make it more likely to require an intervention...surely it would reduce the chance? It looks like Jellyfin/Plex and Emby all support these IDs in the folder names. This is done to remove ambiguity for them, why would it confuse Sonarr? Maybe the text could say it's recommended that you add a media DB ID with IMDB being the preferred one but that the format is media player specific as they all support different formats?

because sonarr uses the show folder to look up the show, and when you do a search on tvdb using `show name (year) imdbid` it will probably give you 0 results.
So I need to find the sweet spot that works with both Sonarr and your used media server software, and I personally use Plex and have actually 0 issues adding tv shows or movies without adding the imdbid.
only happened for me twice and that was for the reboot of magnum and macgyver and that was also because i didn't add the year to the folder.

That's why i suggest it as `optional` over `recommended` so the END user can decide what he wants to use.



> It doesn't, sadly that series is outside Sonarr's control as it doesn't know it exists. It just ignores it, the ID is for the media players benefit. I guess it makes sense therefore to leave it out of this guide. As long as I can get it into the Jellyfin documentation I'm happy with that.

If it's something that Sonarr doesn't handle then theirs no need to add it in a Sonarr suggestion guide and it would fit more in a Jellyfin/Emby guide.

briandipalma:
> because sonarr uses the show folder to look up the show, and when you do a search on tvdb using show name (year) imdbid it will probably give you 0 results.

Would it not parse it based on your naming scheme? If your scheme includes IMDB IDs then it should make it more accurate? It seems like Plex is much better at identifying media, if that's the case then yeah optional makes seems fine.

> If it's something that Sonarr doesn't handle then theirs no need to add it in a Sonarr suggestion guide and it would fit more in a Jellyfin/Emby guide.

Fair.

briandipalma:
I imagine you've seen my comment in the Jellyfin PR but it appears that `[imdbid=tt8001226]` is also supported. Should I raise a new PR to mention that too?

TRaSH-:
image

that's already showing but it would be nicer if jellyfin would support `[imdb-ttxxxxxxx]` over [imdbid-xxxxxxx]
this would make it more compatible for people that like to switch or use both

briandipalma:
I mean `=` instead of `-` e.g. `[imdbid=tt8001226]`. Emby supports it to as it was there before the fork.

TRaSH-:
i'm not planning to add all possible options you can use with plex/emby/jellyfin
being you also could just add the `ttxxxxxxx` 
it's a suggestion that people can use and if possible i try to keep as consistent as possible
my biggest annoyance with the emby/jellyfin option is that it uses/wants (at least from the info i got from you) `imdbid-` vs `imdb-` what we're using for plex.
and the main reason why i think it's annoying is because of compatibility between plex/emby/jellyfin

PR #372, opened 2021-10-13
Remove sudo

sadstan:
It is not advised (and generally not necessary) to run docker as root


TRaSH-:
> It is not advised (and generally not necessary) to run docker as root

true, but in some OS's you won't be able to run those commands without using sudo
and i needed to make it as compatible as possible for all flavor's of OS's

TRaSH-:
that could be a option but this is already added in a note, so you will get a note in a note

PR #387, opened 2021-10-20
NinjaCentral "clean.py" reg-ex fix

ghost:
Hey. Not sure if this is useful for your guide, but I ended up adding this myself, since I found it useful/needed while using NinjaCentral as a nzb indexer.

This is just a small addition to the wonderful `clean.py` script for both NZBGet and SABnzbd.
It adds a simple regex / regex replace line, that helps removing that annoying appending of characters, that NinjaCentral adds to all of their NZBs.

I've used it for quite some time, together with a lot of different NZB indexers, and have observed no negative effects / side effects from using this with other indexers than NinjaCentral.. So it should work without interferring with the other indexers way of doing things.

Again, not sure if this is useful, so you can just reject/close the PR if you want to. Just had this for myself for quite a while, and thought it might be useful for someone else 😄 

Reg-ex test links:
NZBGet:
https://regex101.com/r/KD8uqu/1

SABnzbd:
https://regex101.com/r/fpsQqh/1


VP-EN:
> 
> 
> Thnx for this, are you also on discord ?

Yeah. Same name. Think I'm already in the TrashGuide discord.

PR #426, opened 2021-12-11
Overhaul BD Remux groups

mizz141:
The Current Remux and Blue-Ray groups are very incomplete and sometimes deliver sub-par releases, I have overhauled those with scoring, the list has been made in conjunction with people from the https://releases.moe team, which is a dedicated group focusing on getting the best releases for each anime.


TRaSH-:
> The Current Remux and Blue-Ray groups are very incomplete and sometimes deliver sub-par releases, I have overhauled those with scoring, the list has been made in conjunction with people from the https://releases.moe team, which is a dedicated group focusing on getting the best releases for each anime.

@Visorask can you take a look at it, being it's anime related

Visorask:
@mizz141 Overall I'm not against doing a massive change and I just wanna make sure that changing these won't cause any loops inside Sonarr. So with something like Doki or Judgement will need to have or something similar so if you lose stuff it won't try to upgrade them when re-searching for items. 

```/\b(Doki|-Doki)\b/i```

**Comment by mizz141** [comment]
@Visorask thanks for the reply, however I don't quite understand the issue, but I'll welcome if you do a request change or PR on my fork to illustrate it,
I'll also merge it afterwards.

**Comment by Visorask** [comment]
@mizz141 Unless I've lost my mind, which is possible, it just needs tweaking to have the regex read correctly kind of following my comment and the way the rest currently read. The structure of groups is fine. Don't know if you just kind of followed my regex and imported names. 

Basically, if you need me to, I will clean up the REGEX but the groups I will take your word on as most of my information came from the seadex guys originally when I first started this journey.

**Comment by mizz141** [comment]
@Visorask I'm not experienced with Regex, so I tried what little I could, so getting help with that would be appreciated.

**Comment by bakerboy448** [comment]
`/i` goes at the end of the regex to indicate case insensitive

`/<regexgoeshere>/i`

**Comment by Raventhicc** [comment]
Any update on when can this be implemented?

**Comment by Drapersniper** [comment]
> Any update on when can this be implemented?

just done a PR to @mizz141 branch to fix the regex

**Comment by Raventhicc** [comment]
Awesome, thank you. Just need this to be merged now.

**Comment by TRaSH-** [comment]
@Drapersniper 

Got some errors so it doesn't pass the check.

```
Status: Downloaded newer image for markdownlint/markdownlint:latest
/home/runner/work/Guides/Guides/docs/Sonarr/Sonarr-Release-Profile-RegEx-Anime.md:293: MD009 Trailing spaces
/home/runner/work/Guides/Guides/docs/Sonarr/Sonarr-Release-Profile-RegEx-Anime.md:328: MD031 Fenced code blocks should be surrounded by blank lines
/home/runner/work/Guides/Guides/docs/Sonarr/Sonarr-Release-Profile-RegEx-Anime.md:329: MD031 Fenced code blocks should be surrounded by blank lines
/home/runner/work/Guides/Guides/docs/Sonarr/Sonarr-Release-Profile-RegEx-Anime.md:339: MD031 Fenced code blocks should be surrounded by blank lines
/home/runner/work/Guides/Guides/docs/Sonarr/Sonarr-Release-Profile-RegEx-Anime.md:340: MD031 Fenced code blocks should be surrounded by blank lines

A detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md
```

I recommend to use vscode

Drapersniper:
My bad should have checked the lint before making the original PR, only looked at the regex xD 

The new PR removed the trailing spaces and added empty lines around each code block which should now be passing (don't have docker access or access to markdownlink so can't run it locally, but should have fixed the 3 issues.)

@mizz141 made https://github.com/mizz141/Guides/pull/2 to address @TRaSH- comment.

TRaSH-:
Now we need to wait for @Visorask to click the merge pull button,
being he needs to approve it.
(I don't touch the anime section, only check if it doesn't break anything else)

rg9400:
@mizz141 @Drapersniper @Raventhicc

I think this PR should be reverted and rechecked a bit. Posting my thoughts below which I posted in Discord as well

I think there's a ton of duplicates now. I would revert and ask the OP to make sure they are not double counting groups. There are double counts both in the sections touched by the changes (e.g. NH), as well as double counting because some groups already existed in the sub section like Mysteria, Coalgirls, BlurayDesuYo, etc. Additionally, the OP should check the format of the release groups that are common words in the subs section. Stuff like Judgement or Tsundere needs to have more regex, or else it can match anime titles. Finally, for the groups OP removed completely, like LostYears, Vanilla, SCY, TTGA, BluDragon, ZR, etc., they should list the reasoning. Adding new groups or moving them around is one thing, and though some like McBalls seem a bit questionable as they may be only for one very specific release, I don't know enough to disqualify them. But removing a group requires some explanation imo. Especially if it's because some of those groups are not typically found on stuff like Nyaa but Usenet. I just don't want the OP removing high quality groups just because they aren't released on the places they frequent. If a group is bad, that's another thing, but I think the OP should be clear on why each group was removed from the list completely.

I have a lot of trouble believing stuff like LostYears does not belong in this profile at all.

Visorask:
Now reverted. Follow above advice for any new changes.

rg9400:
To be clear, I think changes are welcome and can improve this profile. I just want to highlight the below points to focus on for the PR

1. **Duplicates**: Make sure there are no duplicates. If a group like GJM appears twice, it will be scored based on each line it appears. This ruins the whole scoring

2. **Formatting**: Be careful with common words. You can see how Visorak uses different regex for different groups. Groups that use common or real words like `Judgement` are dangerous. For example, there is an anime called ` The Seven Deadly Sins: Judgement of Fury`. `\bJudgement\b` would actually match against this, messing up the profile. Instead, use `\[Judgement\]|-Judgement\b` which would only match `[Judgement]` and `-Judgement` in the release title.

3. **Scoring**: I noticed your scoring changes created three tiers, separated by 100. 3900, 4000, 4100. As such, that may be okay. But understand this means that other items can easily cause groups to jump tiers. For example, if an item in the 3900 tier has `Dual.Audio` in the title, it will get an extra +500, causing it to jump past the 4100 tier. Is this what you want? If so, then it is fine. But this is why prior tiers are separated by 500 in most cases

4. **Removed Groups**: As mentioned earlier, I am skeptical of the groups being removed. I know that groups like SCY, ZR, BluDragon, etc. pretty much *only* do Dual Audio releases. By completely removing them, you are essentially saying that an `Erai-raws` release should trump a `LostYears` release. Is this correct? The idea of this profile is to get good subbed releases, but upgrade to dual audio. Maybe LostYears is not the *best* release, but it's certainly better than a sub only release. Of course, if you have logic for why each of these groups should be removed completely, please share. But I think you should move them around in tiers vs removing them completely as these are very consistent dual audio groups that are usually pretty high quality.

5. **Added Groups**: I am not an expert, so I will defer to your judgement on most of these. However, like above, I want to stress this is a dual audio profile, not just a "best anime releases" profile. If a dual audio release is being trumped by a sub only release, I think this is not ideal. There were a few examples why I became skeptical. Like CoalGirls is mainly a fansub group, right? They might have a few dual audio releases, but they are mostly just subs AFAIK. Same with DDY (DameDesuYo/BlurayDesuYo). Another group that made me a bit skeptical was `McBalls` which was placed in the highest tier, but AFAIK, this is a joke tag being used by `Baws` for a Neon Genesis Evangelion release. It's meant to be trumped by a later version of `Arid-Baws`. So seeing it in the highest tier made me a bit worried that similar cases might be happening for other groups. Again, I am no expert, so a lot of these might be justified. Seeing Arid moved up makes sense, and I noticed a bunch of other additions that looked good. I just wanted to highlight that this is being used in automation and cannot be used to target anime by anime specifics, so we need to consider the *average* releases of a group, not just if they have a *single* release where it is the best available option.

mizz141:
Since github is trash and I wasn't pinged on discord, and just found this by accident, here's my answer:

1. Duplicates: GJM must've been an accident on my end since I didn't touch the sub groups, only the BD Groups.

2. Formatting: I'd call this an very isolated incident, if anything, could've been fixed by you with a small PR.

3. Scoring:  Most if not all groups already have all important tags in their respective names, but that is an easy fix. (just make a PR...)

4. Removed Groups: In the original draft SCY was there, I must've made an accident copying them over into final, BluDragon is still a straight no tho... (you also could've just as easily made a PR to re-add SCY in the top tier)

5.  Added Groups: even the joke groups like Balls, not-vodes etc. have the best release, they are just named that way since they were shitty muxes and it's a joke amongst users to do that.

Afterword:

instead of completely removing the whole thing you could've just made a PR fixing some of the underlying issues I have inadvertantly created, but ok I guess.

PR #438, opened 2021-12-20
How to sync 2 radarr sonarr

TRaSH-:
- Created: [NEW] Guide - How to sync 2 radarr/sonarr


bakerboy448:
We should link to this from the radarr and Sonarr /supported wiki pages 

radarr/supported
And
sonarr/supported 

are the paths

TRaSH-:
> We should link to this from the radarr and Sonarr /supported wiki pages
> 
> radarr/supported And sonarr/supported
> 
> are the paths

When it's done we can do this, will also symlink this to the Sonarr tips section 
So it exist in both.

TRaSH-:
Reminder: add info that the sync isn't instant but on a schedule.

PR #485, opened 2022-01-23
Updating Synology.md to remove unnecessary usages of `sudo`

pdcamilleri:
Use of `sudo` to execute `wget` and `mkdir` is not necessary. These are simple commands that do not require elevated permissions to execute correctly.

Also some minor updates to clarify phrasing.


TRaSH-:
@bokkoman & @BaukeZwart could you guys take a look at this, being you maintain the synology guide.
I think there was a reason why we used sudo, or perhaps it's a left over of prior DSM 7.

bokkoman (approved):

PR #552, opened 2022-04-08
Add editorconfig and reformat Sonarr JSON files

rcdailey:
For only the Sonarr JSON files, I performed a "Reformat Document" in Visual Studio Code. This solved a few issues:

- Inconsistent style between JSON files
- Difficulty reading the JSON files in web file viewer due to inconsistent indentation characters
- Lack of consistent style within a single file

There will now be a consistent formatting style going forward enforced by `.editorconfig` file.

The diff is large, but all I did was reformat whitespace. There is NO change to the actual data in the JSON files.

### DIFF
--- a/docs/json/sonarr/.editorconfig
+++ b/docs/json/sonarr/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = false
+
+[*.json]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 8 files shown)


rcdailey:
@austinwbest I know you parse these files with your tooling. I hope the presence of the `.editorconfig` in the `docs/json/sonarr` directory is a non-issue. Can you confirm if that will be ok? I'm hoping you glob for `*.json` and don't just parse any file in the directory regardless of extension. I can move it out if it will be troublesome.

@TRaSH- Let me know if you prefer the `.editorconfig` at the root or at a higher level. The reason why I did not put it higher up is because I am not sure we want the style rules to apply to other files outside of this directory (for example the Radarr CF files)

rcdailey:
@austinwbest When you get some time, could you bless my PR with your approval? Thank you 😃

austinwbest:
Should be fine, i tweaked my side to ignore non `.json` files

PR #555, opened 2022-04-13
added short desc to IMAX

s0up4200:


s0up4200:
Removed due to it not being on a separate branch.

PR #566, opened 2022-04-25
Update requirements.txt

nemchik:
According to https://squidfunk.github.io/mkdocs-material/getting-started/#with-pip 

> This will automatically install compatible versions of all dependencies: [MkDocs](https://www.mkdocs.org/), [Markdown](https://python-markdown.github.io/), [Pygments](https://pygments.org/) and [Python Markdown Extensions](https://facelessuser.github.io/pymdown-extensions/). Material for MkDocs always strives to support the latest versions, so there's no need to install those packages separately.

And we can see https://github.com/squidfunk/mkdocs-material/blob/990b24fc8c38168b3b783bcfd8b551deeed6c0cd/requirements.txt#L21-L27 also includes `mkdocs-material-extensions`


TRaSH-:
thnx

PR #574, opened 2022-05-04
Update "Ignore scene releases" optional

mynameisbogdan:
Update page according to changes in #573.


TRaSH-:
Thnx
I forgot that one being this one only exist for backwards compatibility.

PR #575, opened 2022-05-04
Fix regex word boundary

mynameisbogdan:
Fixed word boundary by removing escaping introduced in #574 by me.

Sorry about that.


TRaSH-:
For json you need the double \

mynameisbogdan:
Indeed, but this is not JSON so it doesn't need escaping. 

My fault is that I copy/pasted from the `optionals.json` in the previous PR.

TRaSH-:
Darn
The JSON is correct
Only on the guide it's wrong.
With Radarr this is just much easier being they officially support import and export of the json

mynameisbogdan:
Yeah, only the guide due to my previous PR. Sorry about that. 🤕

PR #580, opened 2022-05-09
qBittorrent Basic Setup

TRaSH-:
- NEW: qBittorrent Basic Setup


docs/Downloaders/qBittorrent/Basic-Setup.md
```diff
@@ -4,4 +4,251 @@
...
+
+## Downloads
+
+`Tools` => `Options` => `Downloads` (Or click on the cog wheel to access the options)
+
+### When adding a torrent
+
+![!When adding a torrent](images/qb-options-downloads-when-adding-a-torrent.png)
+
+1. For consistent with other torrents I recommend leaving this on `Original`.
```
mynameisbogdan, line 20: Maybe it's just me, but I had some issues with `Original` with torrents that aren't in a subfolder. Sonarr would import subtitles from other torrents, eg. seeing Orange Is The New Black subs in Halo.

TRaSH-: shouldn't be a issue i have a few 100 torrents without a sub folder in my download folder.

Torrents that have multiple files should be in sub folders in the first place, and I never have seen them not in sub folders to be honest.
Are you perhaps using public trackers ?

mynameisbogdan: No, only private trackers. For example Orange is the new black is in subfolder since it was a season pack, but Halo was only the latest episode so a single mkv file. Sonarr would look for srt's in other folders next to halo, and orange is the new black would fit the case.

TRaSH-: sorry never have seen this happen. And if i remember correctly the Starr apps only look in the main folder for the subs and not in the sub folders.
Not sure if Sonarr changed this in the meanwhile.
@bakerboy448 would you know perhaps ?

mynameisbogdan: In my use case it did. I only have "srt" set to auto import. Also I'm 100% certain that in my case it wasn't a manual import, since Halo was only downloaded automatically by Sonarr.

Either way, in my setup `Create subfolder` was a much better setting.

I left a comment on an existing issue on Sonarr but it seems my comment was deleted. Oh well.

TRaSH-: I'm really curious about this, being I've also the .srt set to import.
But to be honest you hardly see them with P2P releases.
Waiting for feedback from @bakerboy448

bakerboy448: This was already fixed in Sonarr.
https://github.com/Sonarr/Sonarr/commit/a6a68b4cae7688506c45ff6cf10989fe6596c274 And https://github.com/Sonarr/Sonarr/commit/d02d1bbdfe8827e13c54ec46aba19339a6fcb17c

original is what should be used

mynameisbogdan: I just have it a test again, and it seems to be fixed. 

Sorry about the noise.

PR #593, opened 2022-05-21
NEW: qBittorrent unRaid Mover

TRaSH-:
- How to run the unRaid mover for qBittorent seeding torrents


docs/Downloaders/qBittorrent/Tips/How-to-run-the-unRaid-mover-for-qBittorrent.md
```diff
@@ -0,0 +1,185 @@
...
+
+Using the following instructions you will be able to move the files with the use of the qBittorrent API.
+
+## Needed
+
+### The Script
+
+Download the following standalone script.
+
+- [Script](https://raw.githubusercontent.com/StuffAnThings/qbit_manage/master/scripts/mover.py){:target="_blank" rel="noopener noreferrer"}
```
mynameisbogdan, line 15: Is mover.old some script we're supposed to download? Or rename mover to mover.old?

TRaSH-: if you use unraid it's already present, or should be present.
haven't checked if they removed it in 6.10

mynameisbogdan: I'm using 6.10 and there is no `mover.old`, only `mover`. Maybe it's for older versions of unraid. But I think it's just a typo in the script.

TRaSH-: image
no typo it exist(ed), but perhaps remove in 6.10

TRaSH-: Also asked someone who has 6.10 running to do a ls
https://cdn.discordapp.com/attachments/840007946545266730/977705366680920134/Screen_Shot_2022-05-21_at_6.51.49_PM.png

mynameisbogdan: For me it's not present on 6.10. 

image

mynameisbogdan: > image
> 
> no typo it exist(ed), but perhaps remove in 6.10
> 
> 

Seems to be from last year, so I don't know it should be really the right mover script to be used that unraid devs intended.

TRaSH-: i asked a few people that run unRaid 6.10 and they all have both.
After some more research and finally just asking @bobokun 

```
Hey there's a advanced mover tuner plugin that modifies what mover does to respect the rules you set so even if you call mover it won't move unless it meets those conditions. I use mover.old so it will move no matter what. It's personal preference
If you don't have the mover tuning plugin just use mover since mover.old will throw an error.
...
Personally I will keep using mover.old but it would error if they don't have the plugin installed since they won't have a mover.old
```

So i'm going to add a extra note for it to the guide thnx

mynameisbogdan: now that makes sense, since I don't have any plugins regarding the mover.

PR #600, opened 2022-05-29
[ANIME] Fix false positive with releases that have E-AC-3 in the name

JackGameer:
To no surprise I came across some badly named anime releases. Falsely matched the AC group regex which resulted in the release having an abnormal high score.

Example release `[naiyas] SPY x FAMILY - S01E04 (WEB 1080p HEVC10 E-AC-3 Multiple Subtitle)`

I'm by no means an expert with regex, I've done some testing and this seems to be fixed now with my tweak.


TRaSH-:
sorry for the delay, i will take a closer look at it this Friday evening or Sunday.
it's kinda busy at my work and hardly have any time

TRaSH-:
@rg9400 could you perhaps take a look at this ?
being i know nothing anime related.

PR #605, opened 2022-06-02
fix: software versions

brettpetch:
Change to more reliable release versioning for versions of deluge and rutorrent.

it should be noted that rutorrent is just a front end and the rtorrent version should be used instead.


TRaSH-:
thnx

PR #613, opened 2022-06-06
Radarr french cf

TRaSH-:
VF: version française (French version)
VOF: version originale française (French original version) 
VFF: version française de France (France French version)
VFQ: version française québécoise (French Canadian version)
VFI: version française internationale (French international version)
VF2: version française double (double French version, should possess both the French and French Canadian audio), which I dunno why sometimes become VF1 or VF3 (I suspect VF1 is only one of the two but unknown and VF3 was with the African French version, only saw this one time) 
TRUEFRENCH: same as VFF
VOSTFR: version originale sous-titré français (original version with French sub)
FANSUB/FASTSUB: sub made by fan (that is how French indexers are reporting it most of the time)

Provided by: [@nicetsy](https://github.com/NiceTSY/)

### DIFF
--- a/docs/json/radarr/user-radarr-cf/.editorconfig
+++ b/docs/json/radarr/user-radarr-cf/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = false
+
+[*.json]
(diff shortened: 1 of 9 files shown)


TRaSH-:
Will be replaced with #645

PR #617, opened 2022-06-08
Update doc for proposed changes to "ignore scene releases" - adds "KOGI"

JonzaUK:
Add's "KOGI" to ignore scene releases to cover Blu renaming the releases.


TRaSH-:
Already covered in #619 
Thnx for reporting.

PR #619, opened 2022-06-08
Added KOGI to the Dislike Scene Releases for indexers that rename

TRaSH-:
- Added: KOGI to the Dislike Scene Releases for indexers that rename
- Added: DEPRECATED warning for the optional.json, please use optionals.json

### DIFF
--- a/docs/Sonarr/Sonarr-Release-Profile-RegEx.md
+++ b/docs/Sonarr/Sonarr-Release-Profile-RegEx.md
@@ -497,7 +497,8 @@ Add this to your `Must not contain (2)`
 Add this to your `Must not contain (2)`
 
 ```bash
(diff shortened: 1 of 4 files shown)


TRaSH-:
I actually wanted to delete it,
It's only used for trash-sync and won't be updated anymore.
Best to upgrade to recyclarr.
But I do except a PR for it.
Not at home anymore and I know I will fail if I try it on my phone.
Thx for reporting

skayo:
> I actually wanted to delete it, It's only used for trash-sync and won't be updated anymore. Best to upgrade to recyclarr. But I do except a PR for it. Not at home anymore and I know I will fail if I try it on my phone. Thx for reporting

The thing is you just caused a bug in recyclarr 😅
recyclarr/recyclarr#87

TRaSH-:
Thnx
Else I would go  extreme and delete it

PR #640, opened 2022-06-24
Radarr CF 20220624

TRaSH-:
- Updated: CF `[LQ]` Separated conditions on request.
- Updated: CF `[HQ-Remux]` Separated conditions on request.
- Updated: CF `[HQ]` Separated conditions on request.
- Updated: CF `[HQ-WEBDL]` Separated conditions on request.
- Updated: CF `[HQ-WEBDL]` added  group `HONE`.
- Updated: CF `[No-RlsGroup]` added group `HONE` to the exemption.

### DIFF
--- a/docs/json/radarr/hq-remux.json
+++ b/docs/json/radarr/hq-remux.json
@@ -5,21 +5,111 @@
   "includeCustomFormatWhenRenaming": false,
   "specifications": [
     {
-      "name": "Remux Groups",
-      "implementation": "ReleaseTitleSpecification",
-      "negate": false,
-      "required": true,
-      "fields": {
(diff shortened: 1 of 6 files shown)


mynameisbogdan:
When it comes to regex, one-liners should be preferred choice. These would be hard to maintain manually...

bakerboy448:
Custom Formats support importing,so manual regex maintenance is moot.

TRaSH-:
> When it comes to regex, one-liners should be preferred choice. These would be hard to maintain manually...

separate conditions are easier to maintain and also easier when you do a manual search and want to see why it matches in the first place.

 
image
*this image is a example test setup, that's why it has a dash at the end*

PR #646, opened 2022-06-29
More release version CFs

NiceTSY:
# Custom release version

Those release version CFs were not included in the base guide. I have found them quite useful.

## SubForced

Indicate subtitles embedded in the video that appear during foreign language passages. HARDSUB is an equivalent.

## FanSub

FanSub, similar to the FastSub tag, indicates that the subtitle is a quick release before the release of an official subtitle.

## Limited

This tag means that the film has had only a limited distribution in cinemas (sometimes less than 250 cinemas). 
Generally, these are art house films, or medium-length films. Similar tags are:
  - Festival: a movie which hasn't been shown in a public theatre, but have been shown on a film festival (such as Cannes Film Festival)
  - STV: Straight to Video, refers to a film that has never been released in theatre
  - DTV: Direct to Video, refers to a film that has never been released in theatre
  - DEV: Directement en vidéo (French for DTV)

## Restored

This tag means that it is an old movie that have been restored to today “standard”.

## HDLight

HDLight, for High Definition Light, refer to a High Definition with a light release weight.
Encoding said to be in HDLight or mHD (medium High Definition), claim to be the best possible resolution while being in a reasonable file size.

## 4KLight

Same as for HDLight but for 4K movies.

## Other notes

md5 ID and score already included.

### DIFF
--- a/docs/json/radarr/user-radarr-cf/movie-version/4klight.json
+++ b/docs/json/radarr/user-radarr-cf/movie-version/4klight.json
@@ -1,0 +1,26 @@
+{
+  "trash_id": "47e3eeee29a787cc09ce0eeedd1d7cb9",
+  "trash_score": "0",
+  "name": "4KLight",
+  "includeCustomFormatWhenRenaming": true,
+  "specifications": [
+    {
+      "name": "2160p",
+      "implementation": "ResolutionSpecification",
+      "negate": false,
+      "required": true,
+      "fields": {
+        "value": 2160
+      }
+    },
+    {
+      "name": "4K Light",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": false,
+      "required": true,
+      "fields": {
+        "value": "\\b4K[ .-]?Light\\b"
+      }
+    }
+  ]
+}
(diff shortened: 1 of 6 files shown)


NiceTSY:
SubForced and FanSub are maybe more related to Anime movies/series or to other language movies which do not have official subs. But the community made some, most of the time it is for Korean/other Asian movies.

LiMiTED is really self-explanatory, it will mostly be seen on arthouse movies.
ex: [French Exit](https://www.themoviedb.org/movie/599960-french-exit) or [Memoria](https://www.themoviedb.org/movie/291874-memoria)

Restored, is for old movies which have been upscaled, or redone for today bigger screens.
ex: [Metropolis](https://www.themoviedb.org/movie/19-metropolis) or [Spartacus](https://www.themoviedb.org/movie/967-spartacus)

Yes, HDLight and 4KLight are definitely LQ (for me at least). But I know more and more people tend to like them. That why it could be interesting to add them. So they can be either scored -10000, 0 (just for tag) or whatever to get them higher.

NiceTSY:
Not stale :)

PR #669, opened 2022-07-17
FonduemangVI\Update hq-web-tier-1

FonduemangVI:
Added ABBiE to Tier 1. @TRaSH- not sure if we should have them in Tier 1 or 2, given SiC and ABBiE are Tier 2 for Sonarr V3. However SiC is Tier 1 for qsonarr

### DIFF
--- a/docs/json/sonarr/v4/normal/hq-web-tier-1.json
+++ b/docs/json/sonarr/v4/normal/hq-web-tier-1.json
@@ -156,13 +156,22 @@
       }
     },
     {
-      "name": "SIC",
+      "name": "SiC",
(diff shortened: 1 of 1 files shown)


TRaSH-:
Hmm, seems i forgot to add this one, being i have it already in mine.
perhaps because i wanted to give it a test first.
eventually i want to have them both in sync with the changes we're doing here, including with Radarr if possible

FonduemangVI:
Ah all good,
I have added it to my Sonarr to test already
That is a good idea

TRaSH-:
closing this for now, being we're going to start the same way we did with Radarr

PR #676, opened 2022-07-25
Add warning to qBittorrent about ZFS/Pre-Allocate

ryanwalder:
ZFS does not support fallocate[1] which means if the `pre-allocate`
option is set to true qBittorrent will error when adding torrent files

1. https://github.com/openzfs/zfs/issues/326

### DIFF
--- a/docs/Downloaders/qBittorrent/Basic-Setup.md
+++ b/docs/Downloaders/qBittorrent/Basic-Setup.md
@@ -29,10 +29,12 @@
 
 1. Pre-allocated disk space for the added torrents, this limits fragmentation and also makes sure if you use a cache drive or a feeder disk that the space is available.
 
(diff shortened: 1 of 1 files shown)


TRaSH-:
> ZFS does not support fallocate[1] which means if the `pre-allocate` option is set to true qBittorrent will error when adding torrent files
> 
> 1. [Support fallocate(2) openzfs/zfs#326](https://github.com/openzfs/zfs/issues/326)

Hopefully I will be home at a reasonable time to review it.
Thnx

TRaSH-:
your changes breaks the markdown

image

I changed it to this

image

Let me know what you think about it, and I'm able to merge it.

ryanwalder:
Ah my bad. Looks good to me!

PR #683, opened 2022-07-28
Removed: markdown pin

TRaSH-:
- Removed: markdown pin


TRaSH-:
Removed because of https://github.com/TRaSH-/Guides/pull/670#issuecomment-1197381318

PR #691, opened 2022-08-01
Remove empty group in optionals

tiemonl:
This empty group caused Recyclarr to fail to update this profile


TRaSH-:
What did I mess up there?
Need to check if I made a mistake and didn't add the new one but edited a existing one

TRaSH-:
I just checked
Seems I screwed up and edited a existing one and didn't add the new one.
I'm not able to fix this on my phone at the moment.
Need the recover the one I edited first and then re-add the new one.

tiemonl:
> I just checked
> 
> Seems I screwed up and edited a existing one and didn't add the new one.
> 
> I'm not able to fix this on my phone at the moment.
> 
> Need the recover the one I edited first and then re-add the new one.

If that is the case, should I close this PR? I see you have the fixes already coming in #692

bakerboy448:
> Need the recover the one I edited first and then re-add the new one.

no change was made / nothing old was removed.

this will get superseded by #692 since it has other needed fixes

PR #695, opened 2022-08-04
Radarr CF 20220804

TRaSH-:
- Changed: Raised score for CF `[Framestore]`.
- Added: regex to prevent download loop with bad dual groups. Issue: #685
- Added: `EXTREME` to the CF `[Bad Dual Groups]`. Issue: #694

### DIFF
--- a/docs/json/radarr/framestor.json
+++ b/docs/json/radarr/framestor.json
@@ -1,6 +1,6 @@
 {
   "trash_id": "ff5bc9e8ce91d46c997ca3ac6994d6f8",
-  "trash_score": "51",
+  "trash_score": "101",
   "name": "FraMeSToR",
   "includeCustomFormatWhenRenaming": false,
   "specifications": [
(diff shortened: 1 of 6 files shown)


docs/json/radarr/framestor.json
```diff
@@ -10,7 +10,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "\\bFraMeSToR\\b"
+        "value": "\\bFraMeSToR$"
```
bakerboy448, line 13: needs to be reverted
doesn't match file names, only release names
`Movie.Title-FraMeSToR.mkv` does not match

docs/json/radarr/hq-remux.json
```diff
@@ -19,7 +19,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "FraMeSToR"
+        "value": "FraMeSToR$"
```
bakerboy448, line 22: revert per above

PR #735, opened 2022-08-18
feat(Anime) - Updated Radarr and Sonarr CFs naming and hashes

FonduemangVI:
Updated naming for CFs. 
Removed braces from Sonarr CFs. 
Removed BMF from anime bd tier 7 due to overlap. 
Updated Radarr dual audio CF
Updated hashes for all name changed CFs both Radarr and Sonarr


FonduemangVI:
Added in uncensored CF that was missing for Sonarr

PR #738, opened 2022-08-19
Update naming scheme to support Plex Editions

thatbritguy:
Updated Radarr naming scheme to support Plex Multiple Editions tags and conditional IMDb tags. #723


exentler:
All CF for Special Editions, Criterion Collection and IMAX must also be updated for the new {edition-} tag

thatbritguy:
@TRaSH- looking again at the quote I don't think the original was a partial quote I think it was missing the `.` at the end so I changed it to that instead of the `...`.

TRaSH-:
@rg9400 is doing a small review on it, and making sure if we need to change the CF.
and we're also considering a small change to the order for consistency.

`The Movie Title (2010) {edition-Ultimate Extended Edition} {imdb-tt0066921}[IMAX HYBRID][Bluray-1080p Proper][3D][DV HDR10][DTS 5.1][x264]-EVOLVE`

vs

`The Movie Title (2010) {imdb-tt0066921} {edition-Ultimate Extended Edition} [IMAX HYBRID][Bluray-1080p Proper][3D][DV HDR10][DTS 5.1][x264]-EVOLVE`

thatbritguy:
With the ones I tested the CFs still seem to pick it up okay at least on renames, including the Theatrical Cut CF.

One suggestion I have would be to consider using `{edition-{EDITION TAGS}}` instead of `{edition-{Edition Tags}}` in the naming - Plex displays the edition in uppercase regardless of how it is formatted in the tags.  I've noticed IMAX always becomes Imax in the filename with the current naming and also the new edition tag format adds a little visual clutter. I've been experimenting and to me the `{edition-EXTENDED}`  in uppercase is clearer at a glance than `{edition-Extended}`.

TRaSH-:
> ..... I've noticed IMAX always becomes Imax in the filename with the current naming ....

IMAX is set by Radarr
Imax Enhanced is handled by CF

thatbritguy:
On any IMAX / IMAX Enhanced the CF part of the name is correct but Radarr puts Imax in the edition tags instead of IMAX for some reason (interestingly shows as IMAX in the naming token preview).

TRaSH-:
@thatbritguy 
I've talked to the team and we decided on the following
`The Movie Title (2010) {imdb-tt0066921} {edition-Ultimate Extended Edition} [IMAX HYBRID][Bluray-1080p Proper][3D][DV HDR10][DTS 5.1][x264]-EVOLVE`

because it looks weird,
if you got `DIRECTORS CUT`
And the Imax we're going to propose to Radarr for a change.

Are you willing to do the changes ?

owine:
Radarr PR to hardcode IMAX uppercase for the Edition Tags. https://github.com/Radarr/Radarr/pull/7566

thatbritguy:
@TRaSH- 
No problem sounds good, I've made the change so should be okay now.

That's great getting the IMAX fixed on Radarr - from the preview in the tokens I suspect it was always intended to be caps anyway.

PR #766, opened 2022-08-28
Change hotio plex docker profile directory

TucoBruto:
It's no longer at /appdata/plex/Profiles

# Pull request

**Purpose**
Detailed description why you created this Pull Request.

**Approach**
If this Pull Request is created to solve a issue explain how does this change address the problem?

- [ ] Use github checklists. When solved, check the box and explain the answer.

**Learning**

**Requirements**

- [ ] These changes meet the standards for [contributing](https://github.com/TRaSH-/Guides/blob/master/.github/CONTRIBUTING.md).
- [ ] I have read the [code of conduct](https://github.com/TRaSH-/Guides/blob/master/.github/CODE_OF_CONDUCT.md).

### DIFF
--- a/docs/Plex/profiles/index.md
+++ b/docs/Plex/profiles/index.md
@@ -13,7 +13,7 @@ To make use of the profiles you need to add it in your your plex folder
 
 Example paths:
 
-* Hotio's container: `/appdata/plex/Profiles`
+* Hotio's container: `/app/usr/lib/plexmediaserver/Resources/Profiles`
 * LSIO container: `/appdata/plex/database/Library/Application Support/Plex Media Server/Profiles` or `/appdata/plex/Library/Application Support/Plex Media Server/Profiles`
 
 ## Profiles


TRaSH-:
This won't be a such as easy change then, because that path only exist inside the container image.
so to make this workable you need to put it somewhere else and mount it inside the container when the container starts

espentruls:
unRAID users can easily migrate the files to the correct path by issuing a simple cp command in the Plex docker terminal.

"cp -r /config/Profiles /app/usr/lib/plexmediaserver/Resources"

TRaSH-:
> unRAID users can easily migrate the files to the correct path by issuing a simple cp command in the Plex docker terminal.
> 
> "cp -r /config/Profiles /app/usr/lib/plexmediaserver/Resources"

if you want to do something like this you should do it correctly so it also sticks if you update/remove what ever you decide to do with the container

image

or using the following option

This will only work for containers using `s6 overlay`, recognisable by `ENVIRONMENT` printed at the top of the log when the container starts.

If you have a need to do additional stuff when the container starts or stops, you can mount your script with the volume `/docker/host/my-script.sh:/etc/cont-init.d/99-my-script` to execute your script on container start or `/docker/host/my-script.sh:/etc/cont-finish.d/99-my-script` to execute it when the container stops. An example script can be seen below.

```
#!/command/with-contenv bash

cp -r /config/Profiles /app/usr/lib/plexmediaserver/Resources
```

[Source](https://hotio.dev/faq/#guides)

Disclaimer: This is untested, being i don't have the time to test it now. and in most cases their is no real need to use one of the profiles if you use a shield, the profiles are actually made for older models shield  (before 2019), and only if you have some playback issues with normal media files

PR #960, opened 2022-11-06
Updated broken Link

quorn23:
# Pull request

**Purpose**
In Misc - TRaSH Sync
Changed link to just the TLD, as the profile.php doesn't exists anymore and is only accessible after login anyways.


TRaSH- (approved):
Thx

PR #966, opened 2022-11-12
[Radarr] Matching `How to setup Custom Formats` with flowchart

TRaSH-:


TRaSH-:
This will superseed #904 and close #904

PR #1017, opened 2022-12-11
[Guide] Move naming scheme to json part1

TRaSH-:
- This is part 1 to move the naming scheme to a json file so 3rd party apps can make use of it. (in part 2 it will be also used in the Guide it self)


rcdailey:
Metadata and corresponding schema changes look good to me. Great job!

PR #1026, opened 2022-12-16
Add `Bravia Core` to Radarr Custom Formats Collection

quorn23:
Added Bravia Core to Radarr collection

**Purpose**

Adding the json to the collection visible on the guide.

**Requirements**


TRaSH- (changes requested):
Now it only needs to show up in the guides,
Could you also add it to the includes.
https://github.com/TRaSH-/Guides/blob/master/includes/cf/radarr-streaming-services.md

quorn23:
> Now it only needs to show up in the guides, Could you also add it to the includes. https://github.com/TRaSH-/Guides/blob/master/includes/cf/radarr-streaming-services.md

Added Bravia Core. Currently without a set score, might make sense to consider a score slightly higher than MA?

owine:
Should have the top streaming score IMO. Higher bitrate and better sound.

TRaSH- (approved):

PR #1053, opened 2022-12-31
Updated to match Disney+

FonduemangVI:
Was not matching Disney+

# Pull request

**Purpose**
Update DSNP CF to fix matching issues for Disney+
Example https://regex101.com/r/AdEeCO/1

**Learning**

**Requirements**

### DIFF
--- a/docs/json/sonarr/cf/dsnp.json
+++ b/docs/json/sonarr/cf/dsnp.json
@@ -10,7 +10,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "\\b(dsnp|dsny|disney|Disney\\+)\\b(?=[ ._-]web[ ._-]?(dl|rip)\\b)"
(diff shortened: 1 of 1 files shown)


TRaSH- (commented):
Don't we need to do this also for Radarr ?

FonduemangVI:
> Don't we need to do this also for Radarr ?

Radarrs CF for DSNP is currently done a different way which is why I didn't update it
image

FonduemangVI:
If we want them to be consistent I can update it and the other Service CFs

bakerboy448 (commented):
Should the sonarr CF not match the radarr CF?

The regex for sonarr was built for release profiles and seems to have been blindly converted to a custom format?

TRaSH-:
> Should the sonarr CF not match the radarr CF?
> 
> The regex for sonarr was built for release profiles and seems to have been blindly converted to a custom format?

It was one of the first one we setup
I think we just used the regex we used in Sonarr being that Sonarr had some weird thing with the quality sources.
But I can't remember correctly

PR #1062, opened 2023-01-02
[Guide] Move naming scheme to json part2

TRaSH-:
- Make use of the json files for the naming scheme


TRaSH-:
@FonduemangVI switched the naming scheme of the guide to the json file
could you check if the anime on is still correct for the Starr apps

FonduemangVI (approved):
All looks good to me @TRaSH-

PR #1065, opened 2023-01-05
exclude hdr10+ from DV HDR10 CF

garypiner:
# Pull request

**Purpose**
The "DV HDR10" currently matches with "DV.HDR10" and "DV.HDR10+", but not "DV.HDR10Plus". This is to update the CF to be consistent and only match with "DV.HDR10"

**Approach**
This is the current regex https://regex101.com/r/YQXE5R/1
This is the new regex https://regex101.com/r/KQO62Q/1
It only matches "DV.HDR10"

**Learning**

**Requirements**

### DIFF
--- a/docs/json/radarr/cf/dv-hdr10.json
+++ b/docs/json/radarr/cf/dv-hdr10.json
@@ -11,7 +11,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(HDR(10)?)\\b)(?=.*\\b(DV|DoVi)\\b)"
+        "value": "^(?=.*\\b(HDR(10)?(?!\\+))\\b)(?=.*\\b(DV|DoVi)\\b)"
       }
     },
     {
--- a/docs/json/radarr/cf/dv-hdr10.json
+++ b/docs/json/radarr/cf/dv-hdr10.json
@@ -1,7 +1,7 @@
 {
   "trash_id": "e23edd2482476e595fb990b12e7c609c",
   "trash_score": "1500",
-  "trash_regex": "https://regex101.com/r/pADWJD/5",
+  "trash_regex": "https://regex101.com/r/pADWJD/8",
   "name": "DV HDR10",
   "includeCustomFormatWhenRenaming": false,
   "specifications": [
--- a/docs/json/radarr/cf/hdr10.json
+++ b/docs/json/radarr/cf/hdr10.json
@@ -19,7 +19,7 @@
       "negate": true,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(HDR(10)?)\\b)(?=.*\\b(DV|DoVi)\\b)"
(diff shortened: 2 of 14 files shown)


docs/json/radarr/cf/dv-hdr10.json
```diff
@@ -11,7 +11,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(HDR(10)?)\\b)(?=.*\\b(DV|DoVi)\\b)"
+        "value": "^(?=.*\\b(HDR(10)?(?!\\+))\\b)(?=.*\\b(DV|DoVi)\\b)"
```
bakerboy448, line 14:
```suggestion
        "value": "^(?=.*\\b(HDR(10)?(?!\\+|P(?:lus)?))\\b)(?=.*\\b(DV|DoVi)\\b)"
```
cover Plus too?

docs/json/radarr/cf/hdr10plus-boost.json
```diff
@@ -20,7 +20,7 @@
       "negate": true,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(HDR(10)?)\\b)(?=.*\\b(DV|DoVi)\\b)"
+        "value": "^(?=.*\\b(HDR(10)?(?!\\+))\\b)(?=.*\\b(DV|DoVi)\\b)"
```
bakerboy448, line 23: why is this needed?

"name": "Not HDR10", covers this

docs/json/radarr/cf/hdr10.json
```diff
@@ -19,7 +19,7 @@
       "negate": true,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(HDR(10)?)\\b)(?=.*\\b(DV|DoVi)\\b)"
+        "value": "^(?=.*\\b(HDR(10)?(?!\\+))\\b)(?=.*\\b(DV|DoVi)\\b)"
```
bakerboy448, line 22: why is this needed?

"name": "Not HDR10+" prevents this

docs/json/radarr/cf/dv-hdr10.json
garypiner: It doesn't need to be explicitly included as the word boundary prevents it from being included

docs/json/radarr/cf/hdr10plus-boost.json
garypiner: Just did a search and replace for the current DV HDR10 regex and replaced it in all files

docs/json/radarr/cf/hdr10.json
TRaSH-: we changed the CF a while ago to make it better matching, and to make it easier to update we prefer to keep the conditions consistent.
but we don't want to go fully one liners

TRaSH- (approved):
The Team took a look and it looks good, appreciate the find and the PR
I only updated the regex example

PR #1086, opened 2023-01-10
[Starr] How to setup Language Custom Formats

TRaSH-:


bakerboy448 (approved):
Seems fine.

@austinwbest any issues with those trash Ids being duped like that?  Can those be filtered out on site side?

PR #1094, opened 2023-01-14
Revert "[Sonarr] Updated naming scheme to be consistent with Radarr."

TRaSH-:
Reverts TRaSH-/Guides#1088


TRaSH-:
Reverted because of the weird outcome with Sonarr

PR #1111, opened 2023-01-20
enable opt-in features

saltydk:
PR enables the following previously enabled options that have now become opt-in:
content.code.copy
content.action.edit
content.action.view
https://squidfunk.github.io/mkdocs-material/upgrade/


TRaSH- (approved):
Thnx for this update

PR #1127, opened 2023-01-30
Add TrueNAS Core guide for hardlinks and general setup

datcactus:
# Pull request

**Purpose**
There currently isn't any information in the TRaSH guides involving TrueNAS/ZFS, only Unraid and a couple others.

**Approach**
Created documentation and updated the index file for How-to-setup-for to include the new documentation link.

**Learning**

**Requirements**


SeiyaGame:
@TRaSH- @datcactus 

I take the liberty of answering here because I also have a NAS under Truenas !
There are a lot of things not correct with the documentation, for example:
 - I advise to create a user/group with an ID higher than 5000 to avoid conflicts, and this on the NAS as well as on the VM because the user/group need to have the same ID on both ( To use NFS )
 - **The SMB share must not be used, the hardlinks do not work properly !**
 - Then for the configuration of the NFS service, it is necessary to define correctly the parameters Number of threads otherwise the NFS service has strange behaviors
 - On the NAS and the VM, we need to configure the NFS domain name (No need for LDAP!) (It's in `/etc/idmapd.conf` for Debian)
 - In the NFS share, you must not put `Mapall Group` or `Mapall User` because otherwise it means that all users on the VM will have access to the NFS share ! (And it's not secure at all)
   The only thing you have to do is to define maproot.
 
 etc etc...
 
 @datcactus , I can help you with the documentation if you want

datcactus:
Thanks, @SeiyaGame! I left you a message on Discord so we can tweak this before I update the PR to fix the markdownlint test fail.

TRaSH-:
Thanks guys
Let me know when I can review it.
Keep in mind I can't check if this is the correct way.
Only thing I'm going to do is check for layout issues and how it's setup etc

TRaSH-:
Also changed it to draft
When you guys think it's done your should be able to switch it back from draft so I know I can take a look at it

mvanbaak:
> * On the NAS and the VM, we need to configure the NFS domain name (No need for LDAP!) (It's in `/etc/idmapd.conf` for Debian)

This is only needed for NFSv4, and only if you want to use id mapping

docs/Hardlinks/How-to-setup-for/TrueNAS-Core.md
```diff
@@ -0,0 +1,132 @@
...
+    To get hardlinks and atomic moves working with ZFS you will need to make use of <u>**ONE**</u> dataset with subfolders.
+
+For this example I'm using my own setup and preferred share `Media`. The naming isn't important as host machine mappings can be manipulated as desired.
+
+Go to Storage -> Pools and click on ⋮ on the right side of your root dataset, then select `Add Dataset`.
+
+![truenas-dataset](images/truenas-create-dataset.png)
+
+On the next screen you'll fill out your basic dataset info, such as name and comments. I have a few things grayed out here as the dataset has already been configured, but in this example you'd use `Media` in the Name field and keep everything else at its default (`Inherit`). During share creation, there will be an option at the bottom for `Share Type`. `Generic` is the default, with `SMB` as a selectable option. Since we'll be using NFS, select `Generic`. Optionally, you can click `Advanced Options` and populate `Quota for this dataset` if you want to limit what the dataset displays and will accept when it's mounted. Click `Submit` (mine shows `Save` since this is an existing dataset) once complete.
+
```
mvanbaak, line 26: A couple of things:
- `recordsize` ZFS property. I have mine set to 1M
- `compression` should be set to `zstd` as it's faster and compresses better
- `atime` is best set to `off` as it's pretty useless on media

Wether you want these options set on this specific dataset, or on the pool, is up to you. Personally I have set both `compression` and `atime` properties on the pool and have them as `inherit` in the dataset. the `recordsize` I set per dataset as it's use case specific.

I dont know the details of TrueNAS, but I would also enable encryption.

mvanbaak (commented):
I'm not using NFS or SMB, so I have no opinions there. As a FreeBSD user I left my remarks on the ZFS usage only.

docs/Hardlinks/How-to-setup-for/TrueNAS-Core.md
mvanbaak: encryption on media is indeed ... only for the extra files (srt, nfo, etc)
I mentioned zstd as it is now the default in FreeBSD :)

But thank you, good job!

datcactus: @TRaSH- changes recorded. Hopefully this is the last revision and we can get an initial version up. We can always tweak it later.

TRaSH- (approved):
looks good, no error in the test setup

PR #1140, opened 2023-02-05
Fix HBO/HMAX inconsistencies across Sonarr V3/V4 and Radarr, Fix HBO rlsgrp matching HBO/HMAX Streaming CF

yammes08:
# Pull request

**Purpose**
There are inconsistencies between the HBO and HBO Max custom formats across the Sonarr V3/V4 and Radarr guides, and also the Sonarr V4 HBO CF was matching any release from the HBO rlsgrp.

**Approach**
- Apply consistent regexes across all areas of the guide, including conditionals for matching the streaming service only
- Created a new HBO Radarr CF for matching older HBO (not HMAX) releases
- Update relevant guide pages and recyclarr configs

**Learning**

**Requirements**

### DIFF
--- a/docs/Radarr/Radarr-collection-of-custom-formats.md
+++ b/docs/Radarr/Radarr-collection-of-custom-formats.md
@@ -52,13 +52,14 @@ I also made 3 guides related to this one.
 | [Hybrid](#hybrid)                             | [BR-DISK](#br-disk)   | [Remux Tier 01](#remux-tier-01)           | [Amazon](#amzn)        |
 | [Remaster](#remaster)                         | [LQ](#lq)             | [Remux Tier 02](#remux-tier-02)           | [Apple TV+](#atvp)     |
 | [4K Remaster](#4k-remaster)                   | [3D](#3d)             | [UHD Bluray Tier 01](#uhd-bluray-tier-01) | [Disney+](#dsnp)       |
-| [Special Edition](#special-edition)           | [x265 (HD)](#x265-hd) | [UHD Bluray Tier 02](#uhd-bluray-tier-02) | [HBO Max](#hmax)       |
-| [Criterion Collection](#criterion-collection) |                       | [UHD Bluray Tier 03](#uhd-bluray-tier-03) | [Hulu](#hulu)          |
-| [Masters of Cinema](#masters-of-cinema)       |                       | [HD Bluray Tier 01](#hd-bluray-tier-01)   | [Netflix](#nf)         |
-| [Theatrical Cut](#theatrical-cut)             |                       | [HD Bluray Tier 02](#hd-bluray-tier-02)   | [Peacock TV](#pcok)    |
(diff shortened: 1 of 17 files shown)


owine:
Thanks for kicking ass

TRaSH-:
we could do the same we did with red
this way when hbo is in the title it won't be matched either

```json
{
  "name": "HBO",
  "includeCustomFormatWhenRenaming": true,
  "specifications": [
    {
      "name": "HBO",
      "negate": false,
      "required": true,
      "fields": {
        "value": "\\b(hbo)(?![-_. ]max)\\b[ ._-]web[ ._-]?(dl|rip)?\\b"
      }
    },
    {
      "name": "WEBDL",
      "implementation": "SourceSpecification",
      "negate": false,
      "required": false,
      "fields": {
        "value": 3
      }
    },
    {
      "name": "WEBRIP",
      "implementation": "SourceSpecification",
      "negate": false,
      "required": false,
      "fields": {
        "value": 4
      }
    }
  ]
}
```

And don't we need to do this also with Radarr ?

yammes08:
Good idea to be more consistent, so have swapped the regex out. There isn't an HBO CF for Radarr.

I did also notice the Sonarr V3 has `\\b(hmax|hbom|hbo max)\\b(?=[ ._-]web[ ._-]?(dl|rip)\\b)/i` so at some point we ought to work out which one is the keeper and go with that. Anyway, this one should be ready to go now.

TRaSH-:
ahh, didn't check and didn't know hbo doesn't do movies, if they do then i forgot to add them

yammes08:
A very quick check of their site suggests that you can stream movies on HMAX, so we should porobably do another CF for that. That'll mean a bunch of updates to other pages/ymls though so should probably do a separate PR for that another day!

owine:
Should we just fold the HBO CF into the HMAX? Radarr has only HMAX. They do have different scores maybe because HBO Go (predecessor to HMAX) quality was not great

yammes08:
> Should we just fold the HBO CF into the HMAX? Radarr has only HMAX. They do have different scores maybe because HBO Go (predecessor to HMAX) quality was not great

I think there might be some work to do to tidy all these up:
Sonarr V3 has HBO and HMAX
Sonarr V4 has HBO and HMAX, but the regexes are a bit different
Radarr has HMAX

We'll need to work out which is the 'best' regex for each, and apply them consistently across all three options. I think I'd probaly still recommend merging this PR so the immediate issue is fixed, but I can open another issue to track tidying up the HBO/HMAX stuff overall.

yammes08:
Don't merge this one, I'm going to add the other HBO/HMAX changes

TRaSH- (commented):
I will take a look at it tomorrow if I'm in time, to late to run a test run now.

TRaSH- (approved):
time for breakage

PR #1151, opened 2023-02-13
Changes for proposed recyclarr template sync

yammes08:
To support a proposed new recyclarr feature, config templates have been relocated to a single folder. A new templates.json file has been created, and the metadata.json and metadata schema have been updated with the new variables.

# Pull request

**Purpose**
At present, recyclarr allows users to create a default configuration file. A proposed new feature would allow for users to list any available templates, and generate a fresh config file from one or more available templates. These changes lay the groundwork in the guide to enable building this function in recyclarr

**Approach**
- Relocate all recyclarr config files to a single location
- create a new `templates.json` file to store locations and IDs (names) of all config files
- Update the guide metadata.json and metadata schema files with new variables

**Learning**

**Requirements**


yammes08:
@rcdailey - here's the PR for the new templates feature for recyclarr. Would be good to get your eye on this one in case I missed anything - thanks again for all your help with this so far!

rcdailey:
Thanks @yammes08, I will try to get a review in for you this week.

rcdailey (commented):
Just to make sure I understand how you expect the behavior to work in Recyclarr. My first line of thinking is that all `id` values will be "globally unique". In other words, you wouldn't be able to have a sonarr `sqp-1` because there's already one for Radarr. The user will see that exact name when they list the templates on the command line.

I think I like this because it's simpler and lets you have more control over the naming scheme. If you would like to scope the IDs per service, you could easily set the `id` to `radarr-sqp-1` and that's what the user would need to type to use that template on the CLI.

yammes08:
Indeed - even though the templates are stratified by service, and hopefully the list command would look something like:

Radarr
SQP-1
Hd-Bluray-web

Sonarr
Web-1080p

...I'd still expect each `ID` to be globally unique. If we ended up in a situation where we wanted two `IDs` to have the same name, then I'd update the `IDs` in the json to be prepended with `<service>-`.

rcdailey (commented):
I think my only feedback at this point is the root path. Instead of:

```
docs/recyclarr-configs
```

I'd go with:

```
docs/recyclarr/configs
```

Making `recyclarr` a subdirectory enables more recyclarr-specific stuff to be added in the future. Maybe it won't always have just config templates.

rcdailey:
I should have said: The directory path is *not* something I feel strongly about. It was (and should have been noted as) a nitpick.

The whole idea behind `metadata.json` is that you should be able to easily move the configs later and update the path there and the tools won't care.

I'm happy to keep it how it is. I just shared a future-proofing thought but I'm probably thinking too far ahead.

yammes08:
I've left the folder as it is for now. @TRaSH- I don't know how to resolve the conflicts with `web_1080p_v4.yml` and `web_2160p_v4.yml` - I think it's because I just updated them to add Comedy Central as per the last update on them, but the ones showing here as conflicting will be deleted/moved to the new location anyway. You should be fine to accept the ones in my PR as the ones to go with. If you're happy, this is all ready to merge.

PR #1153, opened 2023-02-15
Added IMAX Enhanced to Recyclarr configs (non SQP)

s0up4200:
# Pull request

**Purpose**
Added IMAX Enhanced to the recyclarr configs for Radarr (non SQP)

**Approach**
1080p WEB-DLs that are labeled `DSNP` and `IMAX` needs the `IMAX Enhanced` custom format to be recognized as `IMAX`. Right now the configs only have the `IMAX`  custom format added, which means it only grabs IMAX releases not containing the word "DSNP" in them. This results in Radarr not upgrading to DSNP IMAX releases unless they leave out DSNP of the release name.

**Requirements**


TRaSH-:
@yammes08 could you check this, i don't use recyclarr, and don't want to mess things up by importing it blindly.

yammes08:
@TRaSH- the configs match the guides, and the guides don't have IMAX Enhanced in them. Do you want IMAX Enhanced as part of the regular guides?

TRaSH-:
ahh true, i forgot that part because it's only in the SQP
hmm, we might add it to the optional then like we have a few other optionals added ?

yammes08:
The difference here is that under SQP, IMAX-E has to be optional so that folks can choose whether it can outscore an encode or remux as the qualities are grouped. Qualities aren't grouped under the regular guide, so IMAX-E could probably go under Movie Versions as in this context it's much the same as regular IMAX. Unless that causes a major headache!

s0up4200:
> The difference here is that under SQP, IMAX-E has to be optional so that folks can choose whether it can outscore an encode or remux as the qualities are grouped. Qualities aren't grouped under the regular guide, so IMAX-E could probably go under Movie Versions as in this context it's much the same as regular IMAX. Unless that causes a major headache!

I agree. IMAX and IMAX Enhanced should be treated as the same thing (for non-SQP at least), because they are in fact the same thing; more picture and less black bars.

TRaSH-:
okay do it (move it to movie version also for the public guide), when i find the time i will update the guide, shouldn't be much work. but I'm a bit to busy with my own server

yammes08:
I figured you might be a bit snowed under 😅 I'll let you know once it's ready to go.

TRaSH- (approved):
hope i fixed it correctly now

PR #1164, opened 2023-02-16
Move recyclarr property up a level

rcdailey:
The `recyclarr` property was under `json_paths` which isn't appropriate for a few reasons:

1. `json_paths` is expected to be for guide-only data (like custom formats).
2. The recyclarr path at the moment is to YAML files, not JSON files.


rcdailey:
Pulling in @austinwbest and @yammes08 for review.

TRaSH- (approved):
Got a green light to merge it.

PR #1171, opened 2023-02-18
feat(starr): rework of HDR metadata CFs

nuxencs:
# Pull request

**Purpose**
Rework the HDR metadata CFs, to prevent false matches from happening.

**Approach**
Rewrote almost all the regex for the HDR metadata CFs to be oneliners, to prevent false matches from happening.

**Requirements**

- [x] Change regex used in the Radarr CFs
- [x] Change regex used in the Sonarr CFs
- [x] Add regex example links

**Requirements**

### DIFF
--- a/docs/json/radarr/cf/dv-fel.json
+++ b/docs/json/radarr/cf/dv-fel.json
@@ -19,7 +19,7 @@
       "negate": false,
       "required": true,
       "fields": {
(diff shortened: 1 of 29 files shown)


TRaSH- (changes requested):
I checked the regex examples and they all look good,
but i think we also need to transfer this to Sonarr v4.
We can use the same regex example for this case.

TRaSH- (approved):

PR #1180, opened 2023-02-20
Updated golden rule to include capital X and H

luukluuk12:
# Pull request

**Purpose**
I've noticed many show releases are listed as X265 or H265 (capital X and H).

**Approach**
I've updated the sonarr golden rule regex to include capital X or H in the filter.
E.g. simply change regex from (x|h) to (x|h|X|H).

This would probably be helpful for Radarr as well, but I don't know where and how to update it there.


bakerboy448:
Invalid; not needed 

`/I` already tells sonarr to treat it case insensitive

PR #1227, opened 2023-03-18
Simplify/bring consistency to Broadcaster regexes

ColinHebert:
# Pull request

**Purpose**
Broadcaster regexes have slightly different formats.

Detailed description why you created this Pull Request.

**Approach**
This PR is bringing consistency the following rules:
- All broadcaster regexes must start/finish with word boundaries (`\b`)
- Only one capturing group must be present, it should capture the full name of the broadcaster, nothing else
- If webrip/webdl is expected, it will follow the format `[ ._-]web(?:dl|rip)?` (separator followed either `web`, `webrip` or `webdl`). Note, if the content is `web-rip` it will be seen as `web` followed by a word boundary
- Avoid lookaround when not necessary

**Learning**

**Requirements**


docs/json/radarr/cf/hbo.json
```diff
@@ -9,7 +9,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "\\b(hbo)(?![ ._-]max)\\b(?=[ ._-]web[ ._-]?(dl|rip)\\b)"
```
ColinHebert, line 12: There was no need to do a negative lookahead on hboxmax, when there is a positive lookahead on webdl/webrip

docs/json/sonarr/cf/french-canalplus.json
```diff
@@ -27,7 +27,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "\\b(C(ANAL)?[ .]?(\\+|Plus))\\b"
+        "value": "\\b(C(?:anal)?[ .]?(?:\\+\\B|Plus))(?:(?<=\\+)\\B|\\b)"
```
ColinHebert, line 30: additional fix, the previous version could not accept `C+` or `Canal.+` because of the trailing `\b`.
`(?:(?<=\+)\B|\b)` states that the regex should end either in a word boundary (if the last character was in `\w`, *or a non word boundary* if the last character seen is a `+`.

It matches with strings such as:

```
Movie 2022 IMAX 2160p Canal plus WEB-DL TrueHD 7.1 Atmos HDR DoVi HEVC-RlsGrp
Movie 2022 IMAX 2160p Canal + WEB-DL TrueHD 7.1 Atmos HDR10 DoVi HEVC-RlsGrp
```

It will not match with strings such as
```
Movie 2022 IMAX 2160p Canal +3 WEB-DL TrueHD 7.1 Atmos DoVi HDR10 HEVC-RlsGrp
Movie 2022 IMAX 2160p Canal Pluss WEB-DL TrueHD 7.1 Atmos DoVi HDR10 HEVC-RlsGrp
```

yammes08:
Something that isn't accounted for here (and actually isn't accounted for in the current set either) are hybrid releases. You'd want something like `\b(hmax|hbom|hbo[ ._-]max)([ ._-]hybrid)?[ ._-]web(?:dl|rip)?\b` to account for these.

ColinHebert:
> Something that isn't accounted for here (and actually isn't accounted for in the current set either) are hybrid releases. You'd want something like `\b(hmax|hbom|hbo[ ._-]max)([ ._-]hybrid)?[ ._-]web(?:dl|rip)?\b` to account for these.

Would it be a case of using `\b(hmax|hbom|hbo[ ._-]max)[ ._-](?:hybrid|web(?:dl|rip)?)\b`
I haven't seen any hybrid release names as of yet

yammes08:
What I'd forgotten was that it's actually one of the guide CFs adding `Hybrid` into the name in this scenario, which would explain why you'd not seen one like this 'in the wild'. I think the suggested changes look ok to me, but I'm not a regex expert so I'll let the gurus look it over 😄

TRaSH-:
After some research and testing the Simplify, isn't simpler when we need to cover stuff like hybrid etc, and adding all the possible variables in to it is just more complicated then we have it now.
We still can review other settings, but as mentioned before, if you want to help out join our discord and we can work together on it.
So for now this will be on hold/closed

PR #1228, opened 2023-03-18
Treat DV+HDR10 as independent expressions

ColinHebert:
# Pull request

**Purpose**
Currently the DV+HDR10 format is treated as a unit rather than a combination of DV and HDR10.
This leads to some other formats excluding DV, then excluding DV+HDR10 (which is pointless).

**Approach**
- When positive matching, split the DV+HDR10 regex in two independent expression (both required).
- When negative matching
  - in DV formats, only look for HDR10 for exclusion
  - in non DV formats, only look for DV for exclusion

**Learning**

**Requirements**

### DIFF
--- a/docs/json/radarr/cf/dv-hdr10.json
+++ b/docs/json/radarr/cf/dv-hdr10.json
@@ -6,12 +6,21 @@
   "includeCustomFormatWhenRenaming": false,
   "specifications": [
     {
-      "name": "DV HDR10",
+      "name": "DV",
       "implementation": "ReleaseTitleSpecification",
       "negate": false,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(DV|DoVi|Dolby[ .]?Vision)\\b)(?=.*\\b(HDR(10)?(?!\\+))\\b)"
+        "value": "\\b(DV|DoVi|Dolby[ .]?Vision)\\b"
+      }
+    },
+    {
+      "name": "HDR10",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": false,
+      "required": true,
+      "fields": {
+        "value": "\\b(HDR(10)?(?!\\+))\\b"
       }
     },
     {
--- a/docs/json/radarr/cf/dv-hlg.json
+++ b/docs/json/radarr/cf/dv-hlg.json
@@ -14,12 +14,12 @@
       }
     },
     {
-      "name": "Not DV HDR10",
+      "name": "Not HDR10",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(DV|DoVi|Dolby[ .]?Vision)\\b)(?=.*\\b(HDR(10)?(?!\\+))\\b)"
+        "value": "\\b(HDR(10)?(?!\\+))\\b"
       }
     },
     {
--- a/docs/json/radarr/cf/hdr10.json
+++ b/docs/json/radarr/cf/hdr10.json
@@ -14,15 +14,6 @@
       }
     },
     {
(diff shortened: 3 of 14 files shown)


docs/json/radarr/cf/dv-hdr10.json
```diff
@@ -6,12 +6,21 @@
   "includeCustomFormatWhenRenaming": false,
   "specifications": [
     {
-      "name": "DV HDR10",
+      "name": "DV",
       "implementation": "ReleaseTitleSpecification",
       "negate": false,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(DV|DoVi|Dolby[ .]?Vision)\\b)(?=.*\\b(HDR(10)?(?!\\+))\\b)"
```
ColinHebert, line 14: Rather than a double lookahead, match two specifications, DV and HDR10

docs/json/radarr/cf/dv-hlg.json
```diff
@@ -14,12 +14,12 @@
...
     },
     {
-      "name": "Not DV HDR10",
+      "name": "Not HDR10",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
       "required": true,
       "fields": {
-        "value": "^(?=.*\\b(DV|DoVi|Dolby[ .]?Vision)\\b)(?=.*\\b(HDR(10)?(?!\\+))\\b)"
+        "value": "\\b(HDR(10)?(?!\\+))\\b"
```
ColinHebert, line 22: Will in a different PR change the regex for HDR10 to be consistent across all formats

docs/json/radarr/cf/hdr10.json
```diff
@@ -13,15 +13,6 @@
         "value": "\\bHDR10(\\b[^+|Plus])"
       }
     },
-    {
-      "name": "Not DV HDR10",
```
ColinHebert, line 17: The format already excludes DV, no point in excluding DV+HDR10 as well

docs/json/radarr/cf/dv-hdr10.json
TRaSH-: There is a reason why we have them separate
1. The releases often use DV even that the release has DV HDR10
2. When Radarr probes it it will use DV HDR10, so with the change you've done (just did a quick check on my phone) it could result in a download loop

ColinHebert: I'm not 100% sure I follow. The case you're mentioning is
> The release title specifies `DV` but the content is actually `DV HDR10`; radarr probles and puts `DV HDR10` in the filename (on rename). At the end of the operation the file is flagged as `DV HDR10` but shouldn't?

The other part I'm not sure I understand is how materially different is the lookup of both `DV` and `HDR10` in a single specification versus in two specifications.
Both of them are required; the behaviour should be identical.

TRaSH-: the line comparison on github looked a but weird on my phone, and it looked like if you renamed line 5 from the screenshot
image

TRaSH-: > Rather than a double lookahead, match two specifications, DV and HDR10

there are several ways to approach the same result, you can add them all separate you can combine them as 1 liner, you can even setup the regex differently to gain the same results.
i'm sure we had them all as separate conditions, but in the past time we where trying to work around a issue what i personally think is a Radarr/Sonarr v4 glitch, and we won't be able to fix our self.

We're always open for improvement and any help possible, normally we discuss issues in our Private Discord dev support channel. because it's much faster to interact then on github.

ColinHebert: Happy to be added in if it helps with the conversation!

PR #1240, opened 2023-03-27
(update) SABnzdb Scripts compatibility update for V4.0.0 and higher.

quorn23:
# Pull request

**Purpose**
Sabnzbd Version 4.0.0 (current beta) introduced a couple of input variables, that weren't available in the versions before.
This PR fixes the two scripts in the Guide, so they are compatible with earlier and V4+ versions

**Approach**
[https://sabnzbd.org/wiki/scripts/pre-queue-scripts#toc3](https://sabnzbd.org/wiki/scripts/pre-queue-scripts#toc3)

**Requirements**

- [ ] These changes meet the standards for [contributing](https://github.com/TRaSH-/Guides/blob/master/.github/CONTRIBUTING.md).
- [ ] I have read the [code of conduct](https://github.com/TRaSH-/Guides/blob/master/.github/CODE_OF_CONDUCT.md).

### DIFF
--- a/docs/Downloaders/SABnzbd/scripts/Clean/Clean.py
+++ b/docs/Downloaders/SABnzbd/scripts/Clean/Clean.py
@@ -21,8 +21,12 @@ import sys
 import re
 
 try:
-    (scriptname, nzbname, postprocflags, category, script, prio, downloadsize, grouplist, showname, season, episodenumber, episodename) = sys.argv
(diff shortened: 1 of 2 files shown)


TRaSH- (approved):
didn't test it, but i think it's good

fabricionaweb:
I think this broke the script. There is an extra `try`, which is fine just remove it.
But also the params changed, so `directory` variable does not exist. 
Im not sure about how to fix it yet

PR #1424, opened 2023-06-29
ci: Fix markdown lint

rcdailey:
Commit fb6b49f9 introduced a regression in the markdown lint workflow by converting the inline configuration options to a `.markdownlint.jsonc` file. However, this file is not used by the Ruby version of markdownlint (which is used here). It is used by the Node version instead. The correct file would have been `.mdlrc`.

This commit simply restores the inline options as they were before that job was refactored out into its own file. This gets the lint logic working again without much fuss.

### DIFF
--- a/.github/workflows/lint-markdown.yml
+++ b/.github/workflows/lint-markdown.yml
@@ -12,5 +12,5 @@ jobs:
           docker run --rm \
             -v "${GITHUB_WORKSPACE}":/data \
             markdownlint/markdownlint \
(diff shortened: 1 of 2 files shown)


bakerboy448 (changes requested):
Keep the markdown lint file as it's used for VSC and other editors

bakerboy448:
Nvm good as is. Saw the discord discussion. Well add back the jsonc later

bakerboy448 (approved):

PR #1450, opened 2023-07-15
fix(starr): Update EZTV regex

kdbruin:
# Pull Request

## Purpose

Fix the EZTV regex to also match a dot or dash instead of a space character.

## Approach

Replaced the space in the regex to `[- .]` to allow for space, dot and dash.

## Requirements

- [X] These changes meet the standards for [contributing](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CONTRIBUTING.md).
- [X] I have read the [code of conduct](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CODE_OF_CONDUCT.md).


nuxencs (changes requested):
would rather use `[ ._-]` instead of `[- .]` to cover all bases and be consistent with how we account for ` ._-` in other regex examples

kdbruin:
Updated accordingly

nuxencs (approved):

PR #1494, opened 2023-08-08
new(guide): Add mkdocs install & preview instructions to CONTRIBUTING.md

rcdailey:
* Move CONTRIBUTING.md to repo root

  This is the typical location a CONTRIBUTING file is expected to be.

* Add mkdocs install & preview instructions to CONTRIBUTING.md


CONTRIBUTING.md
```diff
@@ -93,3 +93,39 @@ When updating or adding a new CF the test case url needs to be added to the json
 ## Recommendations
 
 - use [VSCode](https://code.visualstudio.com/) for editing with the following extension [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint)
+
+## Preview Docs Locally
+
+### Prerequisites
+
+- Install Python 3
```
bakerboy448, line 101:
```suggestion
- Install [Python 3](https://www.python.org/downloads/)
```

CONTRIBUTING.md
```diff
@@ -93,3 +93,39 @@ When updating or adding a new CF the test case url needs to be added to the json
...
+
+### Mkdocs Installation
+
+First, install mkdocs:
```
bakerboy448, line 105:
```suggestion
First, install [mkdocs](https://www.mkdocs.org/user-guide/installation/):
```

CONTRIBUTING.md
```diff
@@ -93,3 +93,39 @@ When updating or adding a new CF the test case url needs to be added to the json
...
+
+```bash
+pip install mkdocs
+```
+
+Then, install dependent modules:
+
+```bash
+pip install \
+    mkdocs-material \
```
bakerboy448, line 115: can these be moved to a requirements file?
perhaps the same for mkdocs as well

CONTRIBUTING.md
bakerboy448: image

CONTRIBUTING.md
bakerboy448: which python 3 version?

CONTRIBUTING.md
```diff
@@ -93,3 +93,39 @@ When updating or adding a new CF the test case url needs to be added to the json
 ## Recommendations
 
 - use [VSCode](https://code.visualstudio.com/) for editing with the following extension [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint)
```
bakerboy448, line 95: for this; don't we now have a recommended extensions file?

nemchik (changes requested):

requirements.txt
```diff
@@ -0,0 +1,10 @@
+mkdocs
```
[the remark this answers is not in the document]
rcdailey: I suppose I should admit that despite being familiar with Python, I've never used `requirements.txt` before. In fact, when you mentioned it in Discord I thought maybe you were referring to some documentation file, not a functional list of packages.

I think as long as the `docs/requirements.txt` captures all the modules needed to run `mkdocs serve`, I'm happy to get rid of the one I added in favor of the existing one.

> * so that makes reproducibility impossible

Agreed. However I do not understand why reproducibility is important for a docs repo. Can you elaborate? The cost of putting version numbers here is an additional maintenance burden to upgrade packages manually. Not a huge burden, but one that requires someone's time. So I'd like to understand the justification for it.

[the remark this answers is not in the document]
nemchik: > > * so that makes reproducibility impossible
> 
> Agreed. However I do not understand why reproducibility is important for a docs repo. Can you elaborate? The cost of putting version numbers here is an additional maintenance burden to upgrade packages manually. Not a huge burden, but one that requires someone's time. So I'd like to understand the justification for it.

If you build the docs locally with version XYZ of a dependency and the docs are deployed with version ABC of that same dependency things could render differently due to changes in the dependency.

Maintaining and updating the dependency versions is handled automatically via pull requests from dependabot https://github.com/TRaSH-Guides/Guides/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+ so there is no need for anyone to search for when updates are available, just for someone to review the pull request for any breaking change notes and potentially test building the docs locally before merging the dependency changes.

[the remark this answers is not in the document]
rcdailey: Makes sense. Thank you for explaining. I will push a change to use the existing requirements.txt.

bakerboy448:
needs a rebase

rcdailey:
Needed a merge a week ago ;-)

bakerboy448 (approved):

PR #1629, opened 2023-11-12
Remove all `trash_score` properties from CF json files

rcdailey:
Both Notifiarr and Recyclarr no longer use these values.


rcdailey:
CC: @austinwbest @yammes08 @TRaSH-

yammes08 (approved):
My poor eyes.

PR #1553, opened 2023-09-13
fix(starr): Remove RlsGrps from CF TrueHD and TrueHD Atmos

TRaSH-:
# Pull Request

## Purpose

Fix: #1552 
- #1552 

## Approach

- [x] Remove RlsGrps from Radarr `TrueHD`
- [x]  Remove RlsGrps from Radarr `TrueHD Atmos`
- [x]  Remove RlsGrps from Radarr `ATMOS (undefined)`
- [x] Remove RlsGrps from Sonarr `TrueHD`
- [x]  Remove RlsGrps from Sonarr `TrueHD Atmos`
- [x]  Remove RlsGrps from Sonarr `ATMOS (undefined)`

## Open Questions and Pre-Merge TODOs

- [x] Test changes in local test setup

## Requirements

- [x] These changes meet the standards for [contributing](https://github.com/TRaSH-Guides/Guides/blob/master/CONTRIBUTING.md).
- [x] I have read the [code of conduct](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CODE_OF_CONDUCT.md).

PR #1643, opened 2023-11-16
fix(starr): Custom Format Conditional Changes

yammes08:
# Pull Request

## Purpose

- Fix: #1636 
- Partial reversion of #1553 for Radarr - allows TrueHD Atmos matches for groups who don't use full TrueHD Atmos text in releases

## Approach

- [x] Removed `HDR` and `Not SDR` conditionals from 2160p custom formats
- [x] Added release group conditionals back to TrueHD Atmos custom format for Radarr only

- [ ] Regex review to confirm this is the best approach, specifically that -RlsGrp is better in this context

## Requirements

### DIFF
--- a/docs/json/radarr/cf/atmos-undefined.json
+++ b/docs/json/radarr/cf/atmos-undefined.json
@@ -72,6 +72,15 @@
       }
     },
     {
+      "name": "Not RlsGrp (Atmos Only)",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": true,
+      "required": true,
+      "fields": {
+        "value": "W4NK3R|HQMUX"
+      }
+    },
+    {
       "name": "Not TrueHD",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
--- a/docs/json/radarr/cf/atmos-undefined.json
+++ b/docs/json/radarr/cf/atmos-undefined.json
@@ -77,7 +77,7 @@
       "negate": true,
       "required": true,
       "fields": {
-        "value": "W4NK3R|HQMUX"
+        "value": "\\b(W4NK3R|HQMUX)\\b"
       }
     },
     {
--- a/docs/json/radarr/cf/truehd-atmos.json
+++ b/docs/json/radarr/cf/truehd-atmos.json
@@ -23,7 +23,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "\\b(ATMOS|W4NK3R|-DON)(\\b|\\d)"
+        "value": "\\b(ATMOS|W4NK3R|DON)(\\b|\\d)"
       }
     },
     {
--- a/docs/json/radarr/cf/truehd-atmos.json
+++ b/docs/json/radarr/cf/truehd-atmos.json
@@ -14,7 +14,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "True[ .-]?HD"
+        "value": "True[ .-]?HD|W4NK3R|HQMUX"
       }
     },
     {
@@ -23,7 +23,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "\\bATMOS(\\b|\\d)"
+        "value": "\\b(ATMOS|W4NK3R|-DON)(\\b|\\d)"
       }
     },
     {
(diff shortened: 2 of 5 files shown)


yammes08:
@nuxencs - sorry for the tag, but would you mind having a look at this one? Area of focus really is whether the -RlsGrp is the best approach in the context of this custom format to ensure non-matches with other words in the release title.

docs/json/radarr/cf/truehd.json
```diff
@@ -61,6 +61,15 @@
...
         "value": "\\bDD[^a-z+]|(?<!e)ac3"
       }
+    },
+    {
+      "name": "Not RlsGrp (TrueHD only)",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": true,
+      "required": true,
+      "fields": {
+        "value": "CtrlHD|W4NK3R|-DON"
```
nuxencs, line 71:
```suggestion
        "value": "\\b(CtrlHD|W4NK3R|-DON)\\b"
```

docs/json/radarr/cf/atmos-undefined.json
```diff
@@ -71,6 +71,15 @@
         "value": "\\b(l?)PCM(\\b|\\d)"
       }
     },
+    {
+      "name": "Not RlsGrp (Atmos Only)",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": true,
+      "required": true,
+      "fields": {
+        "value": "W4NK3R|HQMUX"
```
nuxencs, line 80: wouldn't `-DON` be needed here as well?
```suggestion
        "value": "\\b(W4NK3R|HQMUX|-DON)\\b"
```

docs/json/radarr/cf/truehd-atmos.json
```diff
@@ -14,7 +14,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "True[ .-]?HD"
+        "value": "True[ .-]?HD|W4NK3R|HQMUX"
```
nuxencs, line 17: wouldn't `-DON` be needed here?

docs/json/radarr/cf/truehd-atmos.json
```diff
@@ -23,7 +23,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "\\bATMOS(\\b|\\d)"
+        "value": "\\b(ATMOS|W4NK3R|-DON)(\\b|\\d)"
```
nuxencs, line 26: wouldn't `HQMUX` be needed here?

nuxencs (dismissed):
some things I would change, maybe I understand it incorrectly, just let me know 😄

docs/json/radarr/cf/truehd.json
yammes08: `\b` means the `-` in `-DON` doesn't match for some reason. Could add the brackets for a capture group though.

docs/json/radarr/cf/atmos-undefined.json
yammes08: I just restored this line back to it's previous state, so I don't think HQMUX is needed.

yammes08: Same issue with the `\b`

yammes08: Sorry - typed the wrong thing hence the comment deletions.

I restored this line to it's previous state. Don't think `-DON` is needed as I don't believe they're naming releases as Atmos undefined when actually they're THD Atmos.

docs/json/radarr/cf/truehd-atmos.json
yammes08: Again, restored this line to it's previous state. This CF needs both `TrueHD` and `ATMOS` conditions to match. So, if a release name has both `TrueHD` and `-DON` in the title then the CF will match, which is what we want to achieve.

docs/json/radarr/cf/truehd-atmos.json
yammes08: Again, restored this line to it's previous state. This CF needs both `TrueHD` and `ATMOS` conditions to match. So, if a release name has both `TrueHD` and `-DON` in the title then the CF will match, which is what we want to achieve.

yammes08:
@nuxencs - have responded to all comments. I don't think we need the changes you've suggested, except perhaps for adding brackets for capturing groups.

nuxencs (approved):
lgtm

PR #1863, opened 2024-03-28
fix(starr): DTS-X CF mismatching

nuxencs:
# Pull Request

## Purpose

Fix the `DTS X` CF mismatching and not matching on various different naming schemes.

## Approach

Update the regex to more accurately match `DTS X`.

## Requirements

### DIFF
--- a/docs/json/radarr/cf/dts-es.json
+++ b/docs/json/radarr/cf/dts-es.json
@@ -58,7 +58,7 @@
           "negate": true,
           "required": true,
           "fields": {
-              "value": "\\b(dts[-_. ]?x)\\b(?!\\d)"
+              "value": "\\b(dts[-_.: ]?x7?)\\b(?![-_. ]?(26[456]))"
           }
       },
       {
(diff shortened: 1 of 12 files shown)


bakerboy448 (commented):
does `(?![-_. ]?(26[456])` need to cover HEVC AVC ect?

nuxencs:
> does `(?![-_. ]?(26[456])` need to cover HEVC AVC ect?

hmm not sure tbh, never seen HEVC or AVC after the audio codec

zakkarry:
> > does `(?![-_. ]?(26[456])` need to cover HEVC AVC ect?
> 
> hmm not sure tbh, never seen HEVC or AVC after the audio codec

search for "DTS HEVC" on bhd yeilds 110 pages.

nuxencs:
> > > does `(?![-_. ]?(26[456])` need to cover HEVC AVC ect?
> > 
> > 
> > hmm not sure tbh, never seen HEVC or AVC after the audio codec
> 
> search for "DTS HEVC" on bhd yeilds 110 pages.

yes but not directly after, which would be the problem here

TRaSH-:
don't we need a updated regex example for future reference, and testing?

TRaSH- (approved):
as discussed in the channel this one was good to go

PR #1874, opened 2024-04-01
fix: anime-dual audio regex.

kaiserbh:
## Purpose

I am encountering issues with Sonarr incorrectly identifying quality upgrades, resulting in false positives. A specific case involved the anime `"Frieren"` where after its final episode aired, the group `LostYears` released a Dual Audio version for episode 26. Although this should have been recognized as a quality upgrade, Sonarr failed to import it automatically. This mishap occurred because the existing regex pattern was only designed to match `"dual-audio"` in lowercase, not accounting for variations where the 'D' and 'A' in "Dual Audio" were capitalized or if there is a space in between of them.

An example would be [this ](https://nyaa.si/view/1797597) release.
Which have the title:
`[LostYears] Frieren Beyond Journey's End - S01E26 (WEB 1080p x265 10-bit AAC E-AC-3) [Dual Audio] | (Sousou no Frieren)`

Previous regex: `dual[ ._-]?audio|[\[(]dual[\])]|(JA|ZH)\+EN|EN\+(JA|ZH)`
Updated regex: `(?:[Dd]ual|[Aa]udio)[ ._-]?(?:[Aa]udio)|[\[(]dual[\])]|(JA|ZH)\+EN|EN\+(JA|ZH)`

## Approach

To resolve this, the regex pattern needs to be adjusted to accommodate both uppercase and lowercase letters for "Dual Audio," as well as to handle potential spaces or hyphens that might appear in the term.

## Requirements

### DIFF
--- a/docs/json/sonarr/cf/anime-dual-audio.json
+++ b/docs/json/sonarr/cf/anime-dual-audio.json
@@ -9,7 +9,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "dual[ ._-]?audio|[\\[(]dual[\\])]|(JA|ZH)\\+EN|EN\\+(JA|ZH)"
+        "value": "(?:[Dd]ual|[Aa]udio)[ ._-]?(?:[Aa]udio)|[\\[(]dual[\\])]|(JA|ZH)\\+EN|EN\\+(JA|ZH)"
       }
     },
     {


bakerboy448:
Presumably invalid. Custom Formats Release Title in Starr already uses case-insensitive regex

kaiserbh:
> Presumably invalid. Custom Formats Release Title in Starr already uses case-insensitive regex

I am not too sure how Sonarr decides what's an upgrade or not as in it can successfully download the file and knows it's an upgrade but doesn't auto import so I assumed because of the regex Release title. 

An example this show is recently downloaded and facing the same issue:
image

It downloaded the dual audio version but doesn't auto import:
image

and below is how I set the upgrades score:
image


So if it's not regex issue it might be just Sonarr issue and I will open an issue there, just making sure it's nothing to do with the custom formats release title.

bakerboy448:
Based on the screenshots the issue is the release title reported from the indexer failing to match the file when ready to import. The item being stuck in activity queue clearly indicates it's not an upgrade and it's easy to see the different CFs between the grab and the on disk file

kaiserbh:
> Based on the screenshots the issue is the release title reported from the indexer failing to match the file when ready to import. The item being stuck in activity queue clearly indicates it's not an upgrade and it's easy to see the different CFs between the grab and the on disk file

Yea okay no, that one makes sense. 

Then what about this?
image

I am assuming it's because of the file name? that it became +600 where it meant to be +2600. 

If so then feel free to close this, as it has nothing to do with the release title regex.

PR #1881, opened 2024-04-13
ci(pr-name-check): integrates pr naming convention into workflow

zakkarry:
# Pull Request

## Purpose

verifies that the naming conventions we use are followed for PR open/reopen/edit

## TODO

- [x] verify that commenting works when pr title is bad
- [x] verify that commenting does not happen when title is good
- [x] verify regex edge cases?


FonduemangVI (approved):
LGTM

PR #1917, opened 2024-05-06
chore(anime): Move Nandesuka to LQ

icco:
# Pull Request

## Purpose

I've regularly seen Nandesuka mislabel and episode numbers. Not positive if that should be cause for moving to LQ, but I find myself regularly blacklisting their releases.

## Requirements

### DIFF
--- a/docs/json/radarr/cf/anime-web-tier-04-official-subs.json
+++ b/docs/json/radarr/cf/anime-web-tier-04-official-subs.json
@@ -97,15 +97,6 @@
       }
     },
     {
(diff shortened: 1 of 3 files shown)


rg9400:
We cannot do this. NanDesuKa is the only group doing a bunch of anime this season. Their quality is also not bad, even though their naming is suspect. I get annoyed by VARGY and NanDesuKa, but they need to remain in the tiers. If you don't grab the shows they are doing, you can move them locally.

icco:
Cool thanks for explaining!

PR #1919, opened 2024-05-07
fix(starr-cf/audio-channels): add word barrier and negates to prevent matching incorrectly

zakkarry:
# Pull Request

## Purpose

adds word barriers after second channel count to prevent matching in cases where numbers in title matches in error

removes 9.1 from guides collections

explicit about negates

## Approach

spoke to @rg9400 and he mentioned this was an issue he had faced. i had the same and this is the solution.

## Requirements


rg9400:
The change should be added to 1.0, 2.0, 3.0, and 4.0 as well. We can delete 9.1 as that isn't a thing. I would also recommend removing some of the negates to match the below screen which is my current setup and should help with some possible loops

image

zakkarry:
> The change should be added to 1.0, 2.0, 3.0, and 4.0 as well. We can delete 9.1 as that isn't a thing. I would also recommend removing some of the negates to match the below screen which is my current setup and should help with some possible loops
> 

I don't know of a situation this would affect the x.0 channel, given that there is no resolution starting with 0 and that is what triggered this false-positive/negate in the regex _FOR ME_, If you are aware of a situation this would happen or think it should be applied universally across them all - i see no downside frankly. :P

updated per your comment, lmk

rg9400:
It's just conservative even though I can't think of any cases either. The changes you made look good to me, but because we are deleting a CF, do changes need to be made on the guide and recyclarr as well?

TRaSH-:
> It's just conservative even though I can't think of any cases either. The changes you made look good to me, but because we are deleting a CF, do changes need to be made on the guide and recyclarr as well?

Only the table and the entry in the collection page

yammes08 (approved):
Think this does it 👍

PR #1923, opened 2024-05-08
fix(guide): unclear qbit mover script host vs url and added cache only

TRaSH-:
# Pull Request

## Purpose

host => The URL you use to access qBittorrent locally. this says to use a URL which would be http://IP:PORT

## Approach

- [x] replace ip with your unraid server ip, replace port with your webgui port you use to access the webgui
- [x] added switch to only pause torrents that are actually on your cache drive

## Requirements


TRaSH-:
> maybe a bit more consistency with the upper and lowercase of `Unraid` 🤔

It kinda sucks how it's used in general. Sometimes you see plain Unraid, other times they use unRaid. To be honest, I don't know what the official naming would be.

zakkarry:
I used to always use unRAID, since RAID is an acronym and all caps are generally used for it....but their site uses Unraid mostly, with a few exceptions - mostly on their forums.

I don't think it really makes much difference, one versus the other, but I think we should try and be somewhat consistent. I like unRAID personally.

nuxencs:
official naming seems to be `Unraid`.

nuxencs (approved):
lgtm

PR #1950, opened 2024-05-22
fix(starr): Moved webdl and webrip into Amazon regex to prevent incorrect matches

wheat4714:
# Pull Request

## Purpose

I created this pull request because I noticed that Amazon CF was applying to certain TV shows and TV episodes. I noticed that the regex is simply looking for "Amazon" in a word boundary. As this is a dictionary word, it's going to match stuff it shouldn't.

## Approach

Instead of matching just "Amazon", it requires additional text either side of, which always comes with these releases. Either the resolution or web-rip / web-dl.

https://regex101.com/r/IR8Qbx/6

## Requirements


wheat4714:
@TRaSH- Better now https://regex101.com/r/IR8Qbx/8

I added a negative lookbehind for "the.amazon" as that seems to be a common pattern, but no sane person would label a streaming service "the streaming service".

I've left all the resolutions in the regex. Removing them would have fixed the incorrect matches as well. Whilst I was not able to find any releases where the resolution is present and WEB-DL/Web-Rip is not, I have only been scrolling through prowlarr results, so cannot be certain that there are none.

nuxencs:
@wheat4714 this is a way simpler version and it still catches 90% of those cases. the last 10% are just horribly named releases.

https://regex101.com/r/kBRsQW/1

nuxencs:
please make the same adjustment for the radarr cf

nuxencs (changes requested):
please make the same adjustment for the radarr cf

nuxencs (changes requested):
please also add back the removed `WEBDL` and `WEBRIP` conditions

nuxencs (approved):

bakerboy448:
Reverting this. Breaking change that materially breaks sonarr rename functionality for existing episodes.

This was not actually tested in sonarr and only has blind regex performed

PR #1963, opened 2024-06-08
fix(radarr): Update x265.json with (?i) to let the entire regular expression case-insensitive

remisharrock:
sometimes I see H 265 or X 265 so I added this modifier to let the entire regular expression case-insensitive. [xh]: Matches "x" or "h" in either lowercase or uppercase due to the (?i) modifier. it will also match "HEVC" in any combination of upper and lower case letters (e.g., "HEVC", "hevc", "Hevc", "hEVC", etc.).

### DIFF
--- a/docs/json/radarr/cf/x265.json
+++ b/docs/json/radarr/cf/x265.json
@@ -9,7 +9,7 @@
       "negate": false,
       "required": true,
       "fields": {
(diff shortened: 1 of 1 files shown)


remisharrock:
I guess we could do the same in many other regexes ?

bakerboy448:
Invalid and not needed. Custom formats are already case sensitive. This will never be added as it is 100% unecessary

PR #1991, opened 2024-06-28
fix(radarr-sonarr): Replace sonarr and radarr anime CF image + add AV1 to anime-radarr

washedszn:
# Pull Request

## Purpose

Addresses issue #1989 - also discovered that AV1 was missing from the anime radarr CF's, so I've added that too.

## Approach

Replaced the two CF images for sonarr and radarr with images which align with the scores in the CF table. As for adding AV1, I added `anime-radarr` to the AV1 JSON object with a score of `-10000` and added a new line to the CF table `radarr-anime.md` so that AV1 is shown and linked correctly

## Requirements


nuxencs (changes requested):
Changes look good, but could you maybe update docs/Radarr/images/cfa-complete.png and docs/Sonarr/images/cfa-complete.png as well? 😄

washedszn:
> Changes look good, but could you maybe update `docs/Radarr/images/cfa-complete.png` and `docs/Sonarr/images/cfa-complete.png` as well? 😄

Just added them! However it got me thinking, for radarr and sonarr, there are two CF images for when `Dual Audio` or `Uncensored` have different scores. Reckon it's worth grabbing new images for them as well? or maybe removing them as it's a bit repetitive

zakkarry:
It's probably best to keep the image as consistent as we can with the scores, especially at PR time.

It's somewhat iffy on whether they can always be remembered to be up to date with the guide (as a whole) - but if we're updating things and its not spur of the moment, I think it's probably best to try and be as consistent at the PR time as possible....regardless of repetition.

In my opinion, the should align with what the majority of users would be seeing in their instance. If there are multiple options than the "most used" option would probably align best with what should be seen in screenshots.

FonduemangVI (approved):

PR #2025, opened 2024-07-28
update(downloaders/deluge): clarify process of setting up thin-client

zakkarry:
clarification on the thin-client setup requiring installation on a second computer where you will be managing Deluge from


TRaSH- (approved):
Should be clean enough for everyone now, thnx for the fast improvement.

PR #2152, opened 2024-12-04
update(starr/cf): add additional service tag in regex and proof-edit definition

zakkarry:
ran across a tag that was previously not included, it isn't really common or anything but it was easy enough to add, why not

decided to proof edit the service's description while I was at it....

let me know what you think

- [ ] make sure DSNP triggers false positives before merging

### DIFF
--- a/docs/json/radarr/cf/htsr.json
+++ b/docs/json/radarr/cf/htsr.json
@@ -10,7 +10,7 @@
       "negate": false,
       "required": true,
       "fields": {
-        "value": "\\b(HTSR|HS)\\b"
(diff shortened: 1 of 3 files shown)


TRaSH- (approved):
It looks good to me, and thanks. As you know, the original description is mainly a copy/paste from the wiki. Thanks for your description.

zakkarry:
> It looks good to me, and thanks. As you know, the original description is mainly a copy/paste from the wiki. Thanks for your description.

Yea, you know me and language and descriptions....i went to add the CF and just thought I'd be thorough.

PR #2153, opened 2024-12-04
feat(starr): Add DiRFiX in repack/proper CF

SeiyaGame:
# Pull Request

## Purpose

French trackers use DIRFIX a lot and mean : 
- Term used when a typo or inaccuracy is detected in the original version name

I don't know if it belongs in this CF because technically the release (the source file) hasn't changed, just a fix has been made to the realease name

## Requirements

- [X] These changes meet the standards for [contributing](https://github.com/TRaSH-Guides/Guides/blob/master/CONTRIBUTING.md).
- [X] I have read the [code of conduct](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CODE_OF_CONDUCT.md).

### DIFF
--- a/docs/Radarr/Radarr-collection-of-custom-formats.md
+++ b/docs/Radarr/Radarr-collection-of-custom-formats.md
@@ -112,6 +112,7 @@ We've made 3 guides related to this.
 | [Multi](#multi)                        |                      |                                                 |
 | [No-RlsGroup](#no-rlsgroup)            |                      |                                                 |
 | [Obfuscated](#obfuscated)              |                      |                                                 |
+| [DiRFiX](#dirfix)                      |                      |                                                 |
 | [Repack/Proper](#repackproper)         |                      |                                                 |
 | [Repack2](#repack2)                    |                      |                                                 |
 | [Repack3](#repack3)                    |                      |                                                 |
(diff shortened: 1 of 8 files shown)


bakerboy448:
This will cause download loops as DIRFIX is not a recognized term for proper in Starr

rg9400:
> This will cause download loops as DIRFIX is not a recognized term for proper in Starr

Right, it needs to be in a separate CF with rename enabled. Alternatively, we can enable rename in this CF itself since we are currently losing track of Repack vs Proper, but this will cause a lot of renames for people. The easier lift is just a separate CF

SeiyaGame:
> Right, it needs to be in a separate CF with rename enabled. Alternatively, we can enable rename in this CF itself since we are currently losing track of Repack vs Proper, but this will cause a lot of renames for people. The easier lift is just a separate CF

I understand and it makes sense, we'll avoid impacting everyone. I'll make a separate CF and score it same as the CF repack/proper

SeiyaGame:
what score would be ideal for this CF?
I scored same as the CF repack/proper but I don't know if it's right.
Maybe put it at `4`? just below repack/proper CF ?

SeiyaGame:
Sorry I renamed the branch on my repo but I didn't know it would close the exit

yammes08:
@NiceTSY - is this something that should be added to the French guide pages?

yammes08:
@SeiyaGame - you will need to rebase and force push to resolve the PR conflicts.

PR #2216, opened 2025-01-12
fix(starr): add MultiSub and Multiple Subtitle to Anime Web Tier 04 Custom Format

dzaqwan:
# Pull Request

## Purpose

Sometimes starr apps fetch release from Erai-raws that is raw or does not have English subtitle

## Approach

Add regex lookaround for `[Multiple Subtitle]` or `[MultiSub]`

image

None

## Requirements

### DIFF
--- a/docs/json/radarr/cf/anime-web-tier-04-official-subs.json
+++ b/docs/json/radarr/cf/anime-web-tier-04-official-subs.json
@@ -39,7 +39,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "\\b(Erai-raws)\\b"
+        "value": "\\b(Erai-raws)\\b(?=.*\\[(MultiSub|Multiple Subtitle)\\])"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


rg9400:
Unforunately, we had to roll back a similar approach we used to do for VARYG. First, there are a variety of subbed Erai-Raws releases that do not contain this string. Just checking the releases from the last 2-3 days and I saw Detective Conan and Beheneko and One Piece Log all missing this string. Next, there will be a variety of reuploads where they strip out this string from the naming, preventing it from matching. Finally, with season packs, this string might not be in the filename always even if it is in the release name, causing download loops.

Trust me, I wanted to do the same thing with VARYG, but it resulted in a ton of support and questions, which is why we removed it. Instead, it is much easier to tell people to ensure their Nyaa is set to the English-translated category and deal with the one-off rare cases where these groups upload raws. Erai has 94 raw releases, an incredibly tiny percentage of their releases, meaning I do not think it is worth solving for those cases and better to just manually intervene when those happen. In terms of support, there has been close to no help needed with VARYG after the rollback whereas when it required this type of string, it caused issues almost every day.

dzaqwan:
@rg9400 Ah somehow I missed it. Thanks for the review anyway!

PR #2240, opened 2025-01-26
fix(Starr): Fix HDR10 regex to correctly exclude +/Plus

yammes08:
# Pull Request

## Purpose

To fix an incorrect regex for HDR10.

## Approach

Use a negative lookahead condition to properly exclude +/Plus from matching

None

## Requirements

### DIFF
--- a/docs/json/radarr/cf/hdr-undefined.json
+++ b/docs/json/radarr/cf/hdr-undefined.json
@@ -39,7 +39,7 @@
       "negate": true,
       "required": true,
       "fields": {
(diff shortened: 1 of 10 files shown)


TRaSH-:
@nuxencs could you take a look at the regex?

nuxencs (approved):
lgtm, but imo we should go over all these metadata CF and standardise them at some point, to make sure they use the same regex when possible

bakerboy448 (approved):

PR #2259, opened 2025-02-07
feat(radarr): Only match B&W edition after year

rg9400:
# Pull Request

## Purpose
The B&W edition CF had a few conditions that matched after year only, but some of the other conditions were generic, causing it to match on movie titles such as the below.

image

## Approach
Extend the regex to match after year to each condition

## Requirements


yammes08 (approved):

nuxencs (approved):
lgtm

PR #2332, opened 2025-03-18
feat(starr-anime): match McBalls for Baws

nuxencs:
# Pull Request

## Purpose

Match `McBalls` as `Baws` as it's just another alias of the same group.

## Approach

Add `McBalls` to the `Baws` conditions.

## Requirements

### DIFF
--- a/docs/json/radarr/cf/anime-bd-tier-02-seadex-muxers.json
+++ b/docs/json/radarr/cf/anime-bd-tier-02-seadex-muxers.json
@@ -75,7 +75,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "\\b(Baws)\\b"
+        "value": "\\b(Baws|McBalls)\\b"
       }
     },
     {
--- a/docs/json/radarr/cf/anime-web-tier-01-muxers.json
+++ b/docs/json/radarr/cf/anime-web-tier-01-muxers.json
@@ -48,7 +48,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "\\b(Baws)\\b"
+        "value": "\\b(Baws|McBalls)\\b"
       }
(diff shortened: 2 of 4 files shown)


rg9400:
Are we sure `McBalls` is the same quality as Baws? IIRC, they used that tag as their "inferior" release and in theory would upgrade to their normal tag if they ever released a proper version of it. I am not sure if that still applies or if they just use random tags for the sake of it.

Regarding the approach, I think regardless we should create a separate condition. Otherwise, it is not easy to tell from a distance if we are capturing that tag or not, only by clicking into this specific condition. It's easier to maintain separate I think, and we follow that with some alternate names for other groups.

nuxencs:
> Are we sure `McBalls` is the same quality as Baws? IIRC, they used that tag as their "inferior" release and in theory would upgrade to their normal tag if they ever released a proper version of it. I am not sure if that still applies or if they just use random tags for the sake of it.

`McBalls` has 2 releases marked as `Best` in the Sneedex.

> Regarding the approach, I think regardless we should create a separate condition. Otherwise, it is not easy to tell from a distance if we are capturing that tag or not, only by clicking into this specific condition. It's easier to maintain separate I think, and we follow that with some alternate names for other groups.

makes sense I guess, just thought that if it's the same group it would make sense to put it in the same definition, but if we already handle that differently, I'm also fine with that

rg9400:
Looks like their initial NGE release was considered "low-effort", but they've done two releases since then under that tag. I am fine adding it to the same tier, and we can reevaluate if they re-release something using the Baws tag to indicate an upgrade.

Regarding as separate conditions, I made that change. Not sure if I should review it since I did the changes, but I will and you can spot check my work and merge

rg9400 (approved):

PR #2343, opened 2025-03-22
feat(starr): Update the DV information for the Infuse player, which appears to support the CMv4.0 metadata

TRaSH-:
# Pull Request

## Purpose

Update the DV information for the Infuse player, which appears to support the CMv4.0 metadata

## Approach

Update the DV information for the Infuse player, which appears to support the CMv4.0 metadata

## Requirements

- [x] These changes meet the standards for [contributing](https://github.com/TRaSH-Guides/Guides/blob/master/CONTRIBUTING.md).
- [x] I have read the [code of conduct](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CODE_OF_CONDUCT.md).

PR #2368, opened 2025-04-14
Revert "feat(starr): Update the DV information for the Infuse player, which appears to support the CMv4.0 metadata"

FonduemangVI:
Reverts TRaSH-Guides/Guides#2343


FonduemangVI:
Revert has not fixed page issue with Discord button

PR #2385, opened 2025-04-27
feat(radarr): Do not match RlsGrp `PiRaTeS` when it is a WEB-DL

TRaSH-:
# Pull Request

## Purpose

After conducting some research, we did not find any issues with their WEBDL, so we decided to remove them from the `LQ` and move them to the `LQ (Release Title)` while excluding them from matching on WEBDL.

The reason they are in `LQ (Release Title)` is that we received several reports indicating that their non-WEBDL versions do not use the best sources (R5) or they use custom discs, resulting in issues such as artifacting in dark scenes, etc.

## Approach

Removed RlsGrp `PiRaTeS` from the CF `LQ` and moved them to the `LQ (Release Title)` while excluding them from matching on WEBDL.

## Requirements

### DIFF
--- a/docs/json/radarr/cf/lq-release-title.json
+++ b/docs/json/radarr/cf/lq-release-title.json
@@ -53,6 +53,15 @@
       }
     },
     {
+      "name": "PiRaTeS (no WEBDL)",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": false,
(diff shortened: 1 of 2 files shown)


nuxencs (approved):
lgtm

yammes08 (approved):

PR #2400, opened 2025-05-25
fix(starr-german): add ML to German CF to avoid matching ML

MarcoCoreDuo:
# Pull Request

## Purpose
When a tracker reports a file as german only and the title contains "German.ML" both the German and German DL (undefined) CF are matching. This should not be the case.

## Approach
Change German CF to not match ML.

## Requirements


PCJones (approved):
Looks good to me, this was an oversight as it happens rarely.

PR #2411, opened 2025-06-14
feat(starr-anime): Assume SubsPlease is from CR

rg9400:
# Pull Request

## Purpose
Right now, SubsPlease does not mark their releases as CR despite only ripping from that source. This means that when people upload these releases onto Usenet or other platforms, they assume it is CR and rename it as such, leading to it matching the CR format. This creates a higher score, leading to unnecessary upgrades. 

## Approach
By adding SubsPlease into the CR format, we will apply that extra score from the beginning, meaning those renamed releases no longer have that incorrect higher score. 

## Requirements


rg9400:
Cannot assume SubsPlease is CR as they also do HiDive releases

PR #2417, opened 2025-06-26
feat(starr-anime): Add B00BA to Anime Tier 03

rg9400:
# Pull Request

## Purpose
B00BA now has 37+ best releases on SeaDex and is a remux group, so adding to the tier with other remux groups with a higher number of best releases

## Approach
Same anime regex

## Requirements

### DIFF
--- a/docs/json/radarr/cf/anime-bd-tier-03-seadex-muxers.json
+++ b/docs/json/radarr/cf/anime-bd-tier-03-seadex-muxers.json
@@ -61,6 +61,15 @@
       }
     },
     {
+      "name": "B00BA",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": false,
+      "required": false,
+      "fields": {
+        "value": "\\b(B00BA)\\b"
+      }
+    },
+    {
       "name": "CBT",
       "implementation": "ReleaseTitleSpecification",
       "negate": false,
--- a/docs/json/sonarr/cf/anime-bd-tier-03-seadex-muxers.json
+++ b/docs/json/sonarr/cf/anime-bd-tier-03-seadex-muxers.json
@@ -70,6 +70,15 @@
       }
     },
     {
+      "name": "B00BA",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": false,
+      "required": false,
+      "fields": {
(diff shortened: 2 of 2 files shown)


nuxencs:
do they only do remuxes or encodes as well? 🤔 
because if they only do remuxes, why not add them to the remux tiers? 😄

rg9400:
> do they only do remuxes or encodes as well? 🤔 because if they only do remuxes, why not add them to the remux tiers? 😄

They only do remuxes it seems, but the anime guides don't have a specific tier list just for remuxes. In theory it can be done, but it's a lot of work to break out the groups that do remux only, encodes only, and do both. Right now, the setup is just going off the SeaDex best release system, and since this group often beats other groups within tiers higher/same as remux, I added it alongside other similar remux only groups (e.g. NAN0). 

If someone is able to split up the tiers by encode vs remux, I would be happy to review it. It's just too much for me to do by  myself right now.

nuxencs:
> but the anime guides don't have a specific tier list just for remuxes

what are these then? 🤔 
https://trash-guides.info/Sonarr/sonarr-collection-of-custom-formats/#anime-bd-tier-05-remuxes
https://trash-guides.info/Radarr/Radarr-collection-of-custom-formats/#anime-bd-tier-05-remuxes

rg9400:
> what are these then? 🤔 https://trash-guides.info/Sonarr/sonarr-collection-of-custom-formats/#anime-bd-tier-05-remuxes https://trash-guides.info/Radarr/Radarr-collection-of-custom-formats/#anime-bd-tier-05-remuxes

This was just my initial grouping in that tier. However, now there are remux groups peppered throughout every other tier as well.  Putting them in tier 3 is based on how they've trumped other encodes/remux groups in tiers 4 and 5, but the fact that they get trumped by tier 1 groups still.

What I meant was that there isn't a specific set of tiers just for remux. That would resolve this, by saying which remuxes are better than others but allowing users to exclude all remuxes if they want. Excluding tier 5 isn't enough to avoid remuxes even now.

nuxencs:
hmmm yeah I think we talked about this before, guess we can do it like this for now until we have a proper separation ready

PR #2426, opened 2025-07-13
fix(starr): remove sphd

lucacome:
# Pull Request

## Purpose

I incorrectly assumed the regex was case sensitive, and now it's marking `SPHD` as scene when it should only be `sPHD`.

Since `sPHD` is not really popular, it makes sense to remove it.

## Requirements

## Summary by Sourcery

Bug Fixes:
- Remove the sPHD custom format entry from Radarr and Sonarr scene JSON to prevent uppercase SPHD from being misclassified as scene

### DIFF
--- a/docs/json/radarr/cf/scene.json
+++ b/docs/json/radarr/cf/scene.json
@@ -14,7 +14,7 @@
       "negate": false,
       "required": true,
       "fields": {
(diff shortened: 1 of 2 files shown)


TRaSH- (approved):
Thx 
I already got a report for it, i completely forgot the valid RlsGrp

PR #2443, opened 2025-07-26
fix(Prowlarr) flaresolverr being marked as currently non functional

Kathund:
# Pull Request

## Purpose

At some point the people behind flare solverr marked the issues as solved and after testing a few minutes ago it seamed to work fine

## Requirements

### DIFF
--- a/docs/Prowlarr/prowlarr-setup-flaresolverr.md
+++ b/docs/Prowlarr/prowlarr-setup-flaresolverr.md
@@ -8,8 +8,6 @@
     - A FlareSolverr Proxy will only be used for requests if and only if the Proxy and the Indexer have matching tags
     - A FlareSolverr Proxy configured without any tags or has no indexers with matching tags will be disabled.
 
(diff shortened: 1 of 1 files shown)


TRaSH- (commented):
I don't think it's wise to remove the bug admonition because we know for a fact that whether it works or not is always a coin toss, depending on whether the Cloudflare developer has broken the latest commit. Even the old one worked for some and not for others. Therefore, it might be wiser to add a bug admonition warning stating that it may work for some, and when it does work, we never know how long it will last.

Kathund:
> I don't think it's wise to remove the bug admonition because we know for a fact that whether it works or not is always a coin toss, depending on whether the Cloudflare developer has broken the latest commit. Even the old one worked for some and not for others. Therefore, it might be wiser to add a bug admonition warning stating that it may work for some, and when it does work, we never know how long it will last.

Sorry for only just getting back to this. Do you think it might be better to then refactor the message to say that it might work for some and is very prone to breaking?

TRaSH-:
> Sorry for only just getting back to this. Do you think it might be better to then refactor the message to say that it might work for some and is very prone to breaking?

Yes, that's correct because we know for a fact that whether it works or not is always a coin toss, depending on whether the Cloudflare developer has broken the latest commit. Even the old one worked for some and not for others. Therefore, it might be wiser to add a bug admonition warning stating that it may work for some, and when it does work, we never know how long it will last. So if you're willing to update this PR with the correct info, we can approve it; otherwise, we will close this PR.

TRaSH-:
This PR has been closed because it has been abandoned.

PR #2465, opened 2025-08-16
feat(starr): Add Streaming Service ATV (This is NOT ATVP)

TRaSH-:
# Pull Request

## Purpose

Added Streaming Service ATV (This is NOT ATVP)

- ATVP is Apple Originals.
- ATV is channels, i.e., non-Apple Originals.

It seems to use the same encoding engine as ATVP, but has DNR applied.

- DNR => (dynamic noise reduction) rather visible, losing details in faces, etc

We still need to decide on which scoring to use.

## Approach

Added Streaming Service ATV (This is NOT ATVP)

- [ ] Set ATV scoring for Sonarr

## Requirements

### DIFF
--- a/docs/json/radarr/cf/atv.json
+++ b/docs/json/radarr/cf/atv.json
@@ -1,0 +1,34 @@
+{
+  "trash_id": "d9e511921c8cedc7282e291b0209cdc5",
+  "name": "ATV",
+  "includeCustomFormatWhenRenaming": true,
+  "specifications": [
+    {
+      "name": "Apple TV",
+      "implementation": "ReleaseTitleSpecification",
+      "negate": false,
+      "required": true,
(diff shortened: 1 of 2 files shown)


beschoenen:
I've added ATV locally, because I was missing it, too.
I did however also add `Apple TV` to the regex.

```
\b(atv|Apple TV)\b
```

TRaSH-:
> I've added ATV locally, because I was missing it, too. I did however also add `Apple TV` to the regex.
> 
> ```
> \b(atv|Apple TV)\b
> ```

Sorry for the late reply, I would suggest using `\b(atv|Apple[ .-_]TV)\b`

TRaSH-:
This PR is included in #2508

PR #2478, opened 2025-09-01
fix(starr): audio channel CFs consistency

nuxencs:
# Pull Request

## Purpose

Fix consistency between the Audio Channel CFs for Sonarr and Radarr and fix an issue with missing word boundaries leading to false negatives.

## Approach

Bring all Audio Channel CFs in line with each other between Sonarr and Radarr and add word boundaries where necessary.

## Requirements

### DIFF
--- a/docs/json/radarr/cf/51-surround.json
+++ b/docs/json/radarr/cf/51-surround.json
@@ -13,7 +13,7 @@
       }
     },
     {
-      "name": "6.1 Surround",
+      "name": "Not 6.1 Surround",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
       "required": true,
--- a/docs/json/radarr/cf/71-surround.json
+++ b/docs/json/radarr/cf/71-surround.json
@@ -13,7 +13,7 @@
       }
     },
     {
-      "name": "5.1 Surround",
+      "name": "Not 5.1 Surround",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
       "required": true,
@@ -22,7 +22,7 @@
       }
     },
     {
-      "name": "6.1 Surround",
+      "name": "Not 6.1 Surround",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
       "required": true,
@@ -36,7 +36,7 @@
       "negate": true,
       "required": true,
       "fields": {
-        "value": "(?<!repac)[^0-9][1-4][ .][0-1]|\\b(Stereo|Mono)\\b"
+        "value": "(?<!repac)[^0-9][1-4][ .][0-1]\\b|\\b(Stereo|Mono)\\b"
       }
     }
   ]
--- a/docs/json/sonarr/cf/71-surround.json
+++ b/docs/json/sonarr/cf/71-surround.json
@@ -13,7 +13,7 @@
       }
     },
     {
-      "name": "5.1 Surround",
+      "name": "Not 5.1 Surround",
       "implementation": "ReleaseTitleSpecification",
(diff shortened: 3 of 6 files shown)


rg9400:
@nuxencs the regex changes look good to me. The one thing is that for some reason, Sonarr does not have the 6.1 surround sound negate, while Radarr does. You added it back to Sonarr, but in my instance I do not have this negate on either CF. I had the word boundaries already, so I suspect I had removed the negate for some particular edge case or reason. I don't remember now, probably a case where a movie was titled as 6.1 but was actually 5.1. 

We can add it to both, just wanted to raise that it might be better to remove from Radarr vs add in Sonarr

nuxencs:
Sounds good to me, I'll change that later today.

docs/json/radarr/cf/71-surround.json
```diff
@@ -13,7 +13,7 @@
       }
     },
     {
-      "name": "5.1 Surround",
+      "name": "Not 5.1 Surround",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
       "required": true,
```
rg9400: We should delete the 5.1 negate in the 7.1 CF so that a misnamed file will always match the higher 7.1 CF

docs/json/sonarr/cf/71-surround.json
```diff
@@ -13,7 +13,7 @@
       }
     },
     {
-      "name": "5.1 Surround",
+      "name": "Not 5.1 Surround",
       "implementation": "ReleaseTitleSpecification",
       "negate": true,
       "required": true,
```
rg9400: We should delete the 5.1 negate in the 7.1 CF so that a misnamed file will always match the higher 7.1 CF

docs/json/sonarr/cf/71-surround.json
```diff
@@ -22,7 +22,7 @@
       }
     },
     {
-      "name": "6.1 Surround",
+      "name": "Not 6.1 Surround",
```
rg9400: The 6.1 negate in the 7.1 CF should be removed as well so that the highest value can be scored when dealing with misnamed files

docs/json/radarr/cf/71-surround.json
```diff
@@ -22,7 +22,7 @@
       }
     },
     {
-      "name": "6.1 Surround",
+      "name": "Not 6.1 Surround",
```
rg9400: The 6.1 negate in the 7.1 CF should be removed as well so that the highest value can be scored when dealing with misnamed files

docs/json/radarr/cf/51-surround.json
```diff
@@ -13,7 +13,7 @@
       }
     },
     {
-      "name": "6.1 Surround",
+      "name": "Not 6.1 Surround",
       "implementation": "ReleaseTitleSpecification",
```
rg9400, line 17: Keep the 6.1 negate out of the 5.1 CF. I think it can go either way (delete the 5.1 negate in the 6.1CF and add 6.1 negate in 5.1 CF), but my gut says there is a reason it was missing since we started with every negate in every CF. It must have been intentionally removed at some point

rg9400:
Highlighting the end result changes we should make

* Delete 5.1 and 6.1 negates in both 7.1 CFs
* Delete 5.1 negate in both 6.1 CFs
* Add 6.1 negate in both 5.1 CFs
* Delete the low channel count negate in both 5.1 CFs

This will force only the highest channel count CF to be scored each time

rg9400 (approved):

PR #2522, opened 2025-10-20
docs(guide): enhance setup guide with folder structure instructions

karanssh:
Added setup instructions for Linux/MacOS and Windows.

# Pull Request

## Purpose

there is a folder structure mentioned, but no way on how to do it. It's easy enough, but added a generic script that does the work as per the guide 

## Approach

just used a bash one liner that should work across all nix type systems.

## Requirements

### DIFF
--- a/docs/File-and-Folder-Structure/How-to-set-up/index.md
+++ b/docs/File-and-Folder-Structure/How-to-set-up/index.md
@@ -3,10 +3,9 @@
 This section will show several how-to guides for installing the Starr Apps.
 This is all done in a way to achieve [Hardlinks and Instant Moves (Atomic-Moves)](/File-and-Folder-Structure/Hardlinks-and-Instant-Moves/).
 
-
 ## Folder Setup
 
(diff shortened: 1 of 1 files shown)


karanssh:
ahh crap, my MD formatting is messed up. Will fix this patchset.

TRaSH- (changes requested):
I don't think this is the best approach to do this, and I also see a few issues.

If I recall correctly, you're using `data/...` as the path, and this could result in the share being created in the home folder instead of the root folder.

Also, we've organized the different operating systems into their own categories to make it easier for end users.

If we use this in a main or central index, end users might accidentally add it for unRAID or Synology (since they also run on Linux-based operating systems), and then they'll create a huge mess.

## Examples where we're already using this option:

- [unRaid](https://trash-guides.info/File-and-Folder-Structure/How-to-set-up/Unraid/#fastest-way-to-create-the-needed-subfolders)
- [Synology](https://trash-guides.info/File-and-Folder-Structure/How-to-set-up/Synology/#if-you-use-usenet) - This one needs improvement, similar to what we use for unRaid

PR #2629, opened 2026-02-20
feat(starr): wire existing quality-profile and profile-groups schemas into pre-commit

owine:
## Summary
- Wire `quality-profile.schema.json` and `profile-groups.schema.json` into pre-commit with 4 new hooks (radarr/sonarr for each)
- Fix `profile-groups.schema.json`: `profiles` was incorrectly typed as `array` but actual data uses an `object` (map of slug → trash_id)
- Fix `quality-profile.schema.json`: add optional `visible` property used by 2 sonarr French profiles
- Fix 6 French quality profile data files: `"group": "21"` (string) → `"group": 21` (integer)

## Test plan
- [x] `pre-commit run check-jsonschema --all-files` passes for all 7 hooks
- [ ] Verify schema rejects files with wrong `profiles` type
- [ ] Verify schema rejects files with string `group` values

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Wire quality profile and profile group JSON schemas into pre-commit and align existing data and schemas with the expected structure.

New Features:
- Add pre-commit hooks to validate Radarr and Sonarr quality profiles against the shared quality profile schema.
- Add pre-commit hooks to validate Radarr and Sonarr quality profile groups against the profile groups schema.

Bug Fixes:
- Correct the profile groups schema to treat profiles as an object map instead of an array, matching existing data.
- Allow an optional visible property in the quality profile schema to support existing French Sonarr profiles.
- Normalize French Radarr and Sonarr quality profile JSON data to use numeric group values instead of strings.

Build:
- Extend pre-commit configuration to run JSON schema validation on quality profile and profile group files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,28 +21,28 @@ repos:
         args: ["-v", "--schemafile", "schemas/sonarr-cf.schema.json"]
 
       - id: check-jsonschema
-        name: Validate Radarr Quality Profiles
-        files: ^docs/json/radarr/quality-profiles/[^/]+\.json$
+        name: Validate Radarr CF Groups
+        files: ^docs/json/radarr/cf-groups/[^/]+\.json$
         types: [json]
-        args: ["-v", "--schemafile", "schemas/quality-profile.schema.json"]
+        args: ["-v", "--schemafile", "schemas/cf-groups.schema.json"]
 
       - id: check-jsonschema
-        name: Validate Sonarr Quality Profiles
-        files: ^docs/json/sonarr/quality-profiles/[^/]+\.json$
+        name: Validate Sonarr CF Groups
+        files: ^docs/json/sonarr/cf-groups/[^/]+\.json$
         types: [json]
-        args: ["-v", "--schemafile", "schemas/quality-profile.schema.json"]
+        args: ["-v", "--schemafile", "schemas/cf-groups.schema.json"]
 
       - id: check-jsonschema
-        name: Validate Radarr Quality Profile Groups
-        files: ^docs/json/radarr/quality-profile-groups/[^/]+\.json$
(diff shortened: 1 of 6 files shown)


owine:
## Re: hook ID uniqueness

The `id` field in pre-commit config must match a hook defined in the upstream repo's `.pre-commit-hooks.yaml` — it's a reference, not a user-defined identifier. Custom IDs like `check-jsonschema-radarr-quality-profiles` would cause pre-commit to fail because no such hook exists in the `check-jsonschema` repo.

Multiple entries with the same `id` but different `files`/`args`/`name` is the standard pattern. The existing config on `master` already uses `id: check-jsonschema` three times (guide-only, radarr CF, sonarr CF), and they all run correctly. The `name` field differentiates them in output.

## Fixed
- PR title updated to `feat(starr): ...` to pass the naming check workflow.

owine:
> All four new pre-commit hooks reuse the same `id: check-jsonschema` within the same repo block; pre-commit requires hook IDs to be unique per repo, so these should be given distinct IDs (e.g., `check-jsonschema-radarr-quality-profiles`, etc.).

The `id` field in pre-commit config must match a hook defined in the upstream repo's `.pre-commit-hooks.yaml` — it's a reference, not a user-defined identifier. Custom IDs like `check-jsonschema-radarr-quality-profiles` would cause pre-commit to fail because no such hook exists in the `check-jsonschema` repo.

Multiple entries with the same `id` but different `files`/`args`/`name` is the standard pattern. The existing config on `master` already uses `id: check-jsonschema` three times (guide-only, radarr CF, sonarr CF), and they all run correctly. The `name` field differentiates them in output.

Also tightened the `files` regex to explicitly match `.json` extension (5b3df8c).

PR #2681, opened 2026-03-30
feat(starr): Add machine-readable conflict declarations for mutually exclusive CFs

rcdailey:
## Summary

Adds `conflicts.json` files for Radarr and Sonarr that declare which custom formats are mutually exclusive, so sync tools can warn users when they select conflicting CFs.

- New `schemas/conflicts.schema.json` for validation
- New `docs/json/radarr/conflicts.json` and `docs/json/sonarr/conflicts.json` with known conflict groups
- `metadata.json` and `metadata.schema.json` updated with `conflicts` path key for both services

## Known conflicts declared

Each service has two conflict groups:

- SDR vs SDR (no WEBDL)
- x265 (HD) vs x265 (no HDR/DV)

These are the only CF-level mutual exclusions documented in the guides (confirmed by searching all markdown and cf-group descriptions for exclusivity language).

## Structure

Based on a discussion with @austinwbest, conflict entries use the trash_id as the object key with `name` and `desc` properties on the value. This is intended to be a forward-looking pattern that could eventually unify how all resource files in the repo are structured.

```json
{
  "custom_formats": [
    {
      "9c38ebb7384dada637be8899efa68e6f": {
        "name": "SDR",
        "desc": ""
      },
      "25c12f78430a3a23413652cbd1d48d77": {
        "name": "SDR (no WEBDL)",
        "desc": ""
      }
    }
  ]
}
```

- Each conflict group is an object keyed by trash_id, so three-way (or larger) mutual exclusion works without redundant pairwise entries
- `name` is required; `desc` is optional and left empty for maintainers to fill in if they want
- Schema validates keys as 32-char hex via `patternProperties`, enforces `minProperties: 2` per group
- `conflicts` paths in metadata point to files (not directories) since there's one file per service

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,7 @@ Here you will find guidelines for contributing to [TRaSH Guides](https://trash-g
         - [Group Specific settings](#group-specific-settings)
         - [Group Custom Format specific settings](#group-custom-format-specific-settings)
     - [quality-profile-groups](#quality-profile-groups)
+    - [conflicts](#conflicts)
 - [Recommendations](#recommendations)
     - [Preview Docs Locally](#preview-docs-locally)
     - [Preview after Pull Request](#preview-after-pull-request)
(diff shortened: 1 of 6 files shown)


rcdailey:
@austinwbest This is the conflicts.json from the Discord discussion. Would appreciate your eyes on the schema since you originally proposed the concept.

rcdailey:
Updated the structure based on a conversation with @austinwbest. Conflict entries now use the trash_id as the object key instead of a property, with `name` (required) and `desc` (optional, empty for now) on the value object.

```json
{
  "9c38ebb7384dada637be8899efa68e6f": {
    "name": "SDR",
    "desc": ""
  }
}
```

The idea is for this to be a starting point for a consistent structure across all resource files in the repo. `desc` is there if maintainers want to add explanations per entry, but it's not populated initially since that's their call.

Schema uses `patternProperties` to validate the hex keys and enforce the value shape. Force-pushed to keep the history clean.

austinwbest (approved):

FonduemangVI (approved):

PR #2694, opened 2026-04-04
feat(sonarr-anime): Enable season-pack cf-group for anime

ColinHebert:
# Pull Request

## Purpose

Anime consumers do like season packs too

## Requirements

## Summary by Sourcery

Enhancements:
- Extend season-pack custom format group JSON to include anime-specific configuration.

### DIFF
--- a/docs/json/sonarr/cf-groups/season-packs.json
+++ b/docs/json/sonarr/cf-groups/season-packs.json
@@ -22,7 +22,8 @@
       "[German] HD Remux + WEB": "0dd5f085ed61a1e01f6d347779dfa1bc",
       "[German] UHD Bluray + WEB": "3b0fa37fddaaefc931b75f2889d4b4f5",
       "[German] UHD Bluray + WEB (Alternative)": "7324309a7d1e10dc0dc2cea6c70ed852",
-      "[German] UHD Remux + WEB": "08cececf1840290f6fd490b7d79e8642"
+      "[German] UHD Remux + WEB": "08cececf1840290f6fd490b7d79e8642",
+      "[Anime] Remux-1080p": "20e0fc959f1f1704bed501f23bdae76f"
     }
   }
 }


TRaSH-:
Did you talk this over with the Anime Team on Discord?

rg9400:
> Did you talk this over with the Anime Team on Discord?

No. That said, I am not opposed to adding it, but what is its default score? I actually use it myself but only with a score of 1 (without issue)

TRaSH-:
> > Did you talk this over with the Anime Team on Discord?
> 
> No. That said, I am not opposed to adding it, but what is its default score? I actually use it myself but only with a score of 1 (without issue)

The default score is 10, and we chose 10 so it would override the repacks of single files, because season packs often don't include the repack in the release name if it's only for 1-2 files. If they add it to the season pack name, it's usually because they messed up the season pack itself—such as missing an episode or including the wrong one.

rg9400:
> The default score is 10, and we chose 10 so it would override the repacks of single files, because season packs often don't include the repack in the release name if it's only for 1-2 files. If they add it to the season pack name, it's usually because they messed up the season pack itself—such as missing an episode or including the wrong one.

I'm not sure how it works with p2p groups doing anime, but for anime groups, they do often include it in the name. I think 10 might be too aggressive for our profile scoring. I just haven't tested it

TRaSH-:
> I'm not sure how it works with p2p groups doing anime, but for anime groups, they do often include it in the name. I think 10 might be too aggressive for our profile scoring. I just haven't tested it

Episodes that need a repack will include it in the episode title. After the season has aired, they will release the season pack under the showname.Sxx.blabla (with the repacks included). If they mess up the season pack, they will create a new one and name it something like showname.sxx.repack.blabla.

bakerboy448:
conflicted

PR #2696, opened 2026-04-07
fix(downloaders): ensure fclones remains executable after reboot on unRAID

Pa7rickStar:
# Pull Request

## Purpose

On unRAID, the `fclones` binary is restored to `/usr/local/bin` at boot through the generated `/boot/config/go` boot-time setup. The current script appends `PATH` export and `cp`, but does not also restore the execute bit after the copy. As a result, `fclones` can end up present after reboot but not executable. The file this change touches is `includes/downloaders/mover-tuning-end.sh`.

## Approach

Add the following line to the block that writes the `fclones boot-time setup` section into `/boot/config/go`:

```sh
echo "chmod +x /usr/local/bin/fclones" >> "$GO_FILE"
```
This ensures the copied binary is executable on every boot, not only when the installer is first run. While

```sh
chmod +x "$BOOT_DIR/fclones" 2>/dev/null || true
```
is already applied there are cases where these permissions do not persist. This is a minimal fix for those cases.

None.

## Learning

None.

## Requirements

## Summary by Sourcery

Bug Fixes:
- Restore execute permissions for the fclones binary on unRAID at boot by adding a chmod step to the boot-time setup script.

### DIFF
--- a/includes/downloaders/mover-tuning-end.sh
+++ b/includes/downloaders/mover-tuning-end.sh
@@ -3,8 +3,8 @@ set -euo pipefail # Exit on error, undefined variables, and pipe failures
 
 # =====================================
 # Script: qBittorrent Cache Mover - End
-# Version: 1.3.0
-# Updated: 20260314
+# Version: 1.3.1
+# Updated: 20260411
 # =====================================
 
 # Script version and update check URLs
--- a/includes/downloaders/mover-tuning-end.sh
+++ b/includes/downloaders/mover-tuning-end.sh
@@ -247,6 +247,7 @@ install_fclones_binary() {
                 echo "# fclones boot-time setup" >> "$GO_FILE"
                 echo "export PATH=/usr/local/bin:\$PATH" >> "$GO_FILE"
                 echo "cp $BOOT_DIR/fclones /usr/local/bin/fclones" >> "$GO_FILE"
+                echo "chmod +x /usr/local/bin/fclones" >> "$GO_FILE"
             else
                 log "⚠ Cannot write to $GO_FILE. Please check permissions. (continuing anyway)"
             fi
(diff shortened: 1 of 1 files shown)


TRaSH-:
@Pa7rickStar, could you please review what sourcery-ai suggested [HERE](https://github.com/TRaSH-Guides/Guides/pull/2696#pullrequestreview-4066914728)?

Pa7rickStar:
@TRaSH- Yes, already on it. `install -m 755 "$BOOT_DIR/fclones" /usr/local/bin/fclones` seems like the right approach here.

The simple fix would be to replace:

`echo "cp $BOOT_DIR/fclones /usr/local/bin/fclones" >> "$GO_FILE"`

with something like:

`echo "[ -f \"$BOOT_DIR/fclones\" ] && install -m 755 \"$BOOT_DIR/fclones\" /usr/local/bin/fclones" >> "$GO_FILE"`

However, that would only apply to new installs. In cases like mine, existing users would still need to make `fclones` executable after each reboot.

I already have a solution that checks `/boot/config/go` for an existing `fclones` boot-time setup and updates it accordingly. Since modifying that file is somewhat sensitive, I’d like to test it a bit more first.

Pa7rickStar:
@TRaSH- I tested the migration logic against several mock `/boot/config/go` variants.

The current implementation correctly upgrades the main legacy case (`cp ... /usr/local/bin/fclones`) and preserves surrounding user changes. However, it does not yet handle some existing-marker edge cases, such as incomplete `fclones` blocks, commented-out legacy lines, or pre-existing `install` lines that are not guarded against a missing `fclones` binary.

Those edge cases are most likely the result of manual user changes and cause the installer to leave the `go` file unchanged, which matches the previous behavior. Because of that, I do not think it is necessary to guard against them in this PR.

Here are the [outputs](https://github.com/user-attachments/files/26574174/output_summary.txt) of my tests. The edge cases are `7`, `10` and `12`. In all other cases the update is applied or an already migrated go file is left as is (case `4`).

includes/downloaders/mover-tuning-end.sh
BaukeZwart, line 1: The proposed fix is solid only we also need to change the scripts version number, I propose `Version: 1.3.1`

TRaSH- (approved):
apporved by @BaukeZwart

PR #2738, opened 2026-05-15
chore(gha): Only deploy PR previews when rendered content changes

owine:
## Summary

`Build PR & Deploy` rebuilds the whole mkdocs site and pushes it to Cloudflare on every PR, including ones that change nothing the site renders like workflows, schemas, or scripts. Those runs waste a build and leave a preview comment identical to production.

The decision is split across the two workflows in the chain. `Trigger PR Build` fires on `pull_request`, so it has the diff context; it runs `dorny/paths-filter` (pinned to `v4.0.1`) and writes the result into the existing `pr_info` artifact as `SHOULD_DEPLOY`. `Build PR & Deploy` is triggered by `workflow_run`, where path filters aren't available, so it gets a small `check` job that reads the flag. The `build` and `comment` jobs depend on `check` and only run when the flag is `true`.

The filter matches what feeds the rendered site: `docs/`, `images/`, `includes/`, `overrides/`, and `mkdocs.yml`. `docs/json/**` doesn't need its own entry because the `markdownextradata` plugin pulls it into pages, so `docs/**` already covers it.

Fork protection is unchanged. The `check` job keeps the `fork == false` guard, and because `build` and `comment` read `needs.check.outputs.*`, they skip when `check` is skipped on a fork.

## Why split it this way

`paths:` filters only work on `push` and `pull_request`, not `workflow_run`. Filtering inside `Build PR & Deploy` directly isn't an option, and rebuilding the changed file list there from the `workflow_run` payload is fragile. Deciding once in the workflow that has the diff, and carrying that decision in the artifact, avoids both problems.

## Test plan

- [x] `actionlint` clean on both workflows (its shellcheck integration also surfaced two pre-existing SC2086/SC2129 warnings in deploy-pr.yml, fixed here)
- [x] `pre-commit run` clean locally (yamllint, editorconfig)
- [x] Both workflows parse as valid YAML
- [ ] After merge: a docs-only PR still builds, deploys, and comments
- [ ] After merge: a workflow/schema/script-only PR runs `check`, skips `build` and `comment`, and posts no preview comment

## Summary by Sourcery

Gate PR preview builds and comments on whether the pull request changes rendered documentation content, instead of always deploying on every PR.

New Features:
- Introduce a SHOULD_DEPLOY flag passed between workflows via the existing pr_info artifact to control whether a PR preview is built and deployed.

Enhancements:
- Split the deploy-pr workflow into a check job that reads PR metadata and a build job that only runs when a preview should actually be deployed.
- Update the comment job in the deploy-pr workflow to depend on the check result and use its outputs instead of recomputing PR metadata.
- Pin and configure dorny/paths-filter in the trigger-pr-deploy workflow to detect changes to docs and other rendered site assets before triggering a deploy.
- Clean up shell usage in deploy-pr.yml to avoid unquoted variable expansions and related shellcheck warnings.

CI:
- Adjust GitHub Actions workflows so PR preview builds, deployments, and comments are skipped for changes that do not affect rendered site content.

### DIFF
--- a/.github/workflows/deploy-pr.yml
+++ b/.github/workflows/deploy-pr.yml
@@ -16,12 +16,11 @@ permissions:
   actions: read
 
 jobs:
-  build:
+  check:
     runs-on: ubuntu-latest
     if: github.event.repository.fork == false
     outputs:
-      url: ${{ steps.cloudflare.outputs.deployment-url }}
-      branch_url: ${{ steps.cloudflare.outputs.pages-deployment-alias-url }}
+      should_deploy: ${{ steps.pr_info.outputs.should_deploy }}
       pr_number: ${{ steps.pr_info.outputs.pr_number }}
       pr_sha: ${{ steps.pr_info.outputs.pr_sha }}
     steps:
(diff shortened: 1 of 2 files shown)


owine:
Thanks for the review. Both points considered:

1. SHOULD_DEPLOY resilience: done in 661ece1. The check job now reads it with a fallback and defaults missing or empty to `false`, so an artifact from a Trigger PR Build run that predates the path filter skips the deploy instead of failing the `cat` under `bash -e`.

2. Normalizing the paths-filter output: leaving this out, with reasoning. `dorny/paths-filter` is pinned to a commit SHA (`v4.0.1`), so its `true`/`false` output is frozen for us. It can't change unless we deliberately move the pin, and we'd re-verify the contract at that point. A normalization layer would guard a case the SHA pin already rules out.

bakerboy448 (approved):

PR #2740, opened 2026-05-16
chore(ci): harden GitHub Actions workflows with zizmor

rcdailey:
# Pull Request

## Purpose

This PR adds [zizmor](https://docs.zizmor.sh/) as a pre-commit hook and addresses the security findings it reports against the GitHub Actions workflows in this repo.

zizmor is a static analysis tool for GitHub Actions workflows. It catches common security misconfigurations (credential leaks, template injection, overly broad permissions, etc.) before they ship. Adding it to pre-commit means future workflow changes get checked automatically.

## Approach

**Fixes applied:**

- Added `persist-credentials: false` to checkout steps that don't need git push access. This prevents the GitHub token from being persisted in the local git config, reducing the attack surface if an action later accesses the `.git` directory.
- Added explicit `permissions:` blocks to workflows that previously relied on the default (broad) token permissions. Each job now requests only what it actually needs.
- Moved the `deploy-pr.yml` permissions from workflow-level to job-level so the build job (read-only) doesn't inherit write permissions meant for the comment job.
- Replaced manual `actions/cache` + pip cache directory steps with `setup-python`'s built-in `cache: pip` in the deploy workflows. Same caching behavior, less boilerplate.
- Converted template expressions (`${{ ... }}`) in `run:` blocks to environment variables where possible. This prevents potential code injection if an interpolated value ever contains shell metacharacters. Specifically:
  - `deploy-pr.yml` comment step now uses `env:` block instead of inline interpolation
  - `automatic-changelog.yml` uses `${LAST_EDIT_TIME}` (shell expansion) instead of `${{ env.LAST_EDIT_TIME }}` (template expansion)

**Intentional suppressions (with inline justification comments):**

- `artipacked` on the two checkouts that use `secrets.PAT`: these workflows need credentials persisted to push commits back to the repo.
- `dangerous-triggers` on `deploy-pr.yml`: the `workflow_run` trigger is the standard secure pattern for accessing secrets in PR preview deployments.
- `cache-poisoning` on the deploy workflows: only trusted pushes to master (or repo-owned workflow_run triggers) can write to the cache; fork PRs cannot poison it.

**Pre-commit hook:**

Added zizmor v1.25.0 via the [zizmorcore/zizmor-pre-commit](https://github.com/zizmorcore/zizmor-pre-commit) repo. It runs on `.github/workflows/*.yml` files.

- [ ] Verify the deploy workflows still cache pip correctly with `cache: pip` on `setup-python` (this is functionally equivalent to the old manual approach, just less code)

## Requirements

## Summary by Sourcery

Harden GitHub Actions workflows by tightening permissions, reducing credential exposure, and integrating automated security scanning for workflow configs.

New Features:
- Add zizmor static analysis as a pre-commit hook to automatically scan GitHub Actions workflows for security issues.

Enhancements:
- Add explicit, least-privilege permissions to multiple workflows and jobs that previously relied on default token permissions.
- Disable credential persistence for checkout steps that do not require push access to reduce token exposure risk.
- Simplify Python dependency caching in deploy workflows by using setup-python's built-in pip cache instead of manual cache steps.
- Refine workflow scripting by replacing GitHub expression interpolation in shell commands with environment variables to mitigate potential injection risks.
- Clarify and adjust workflow_run and caching behavior with documented, intentional security rule suppressions where required.
- Set an explicit empty permissions block in the hold-merge workflow to prevent unintended token access.

CI:
- Strengthen CI security posture across workflows by standardizing permissions, token handling, and cache usage.

Deployment:
- Ensure deployment-related workflows request only necessary permissions and use safer token and caching configurations.

### DIFF
--- a/.github/workflows/automatic-changelog.yml
+++ b/.github/workflows/automatic-changelog.yml
@@ -14,7 +14,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout
+      # PAT is required to push commits back to the repository.
+      - name: Checkout # zizmor: ignore[artipacked]
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
@@ -28,7 +29,7 @@ jobs:
 
       - name: Run git log
         run: |
-          git log --no-merges --perl-regexp --author='^(?!.*(\[bot\]|actions@github\.com)).*$' --grep='^(feat|fix|docs|style|refactor|perf|test|build|revert|update)(\((?!mkdocs|pr-naming-check).*?\))?:(?! Update CONTRIBUTORS.md)' --since="${{ env.LAST_EDIT_TIME }}" --pretty="%s" > log_output.txt
+          git log --no-merges --perl-regexp --author='^(?!.*(\[bot\]|actions@github\.com)).*$' --grep='^(feat|fix|docs|style|refactor|perf|test|build|revert|update)(\((?!mkdocs|pr-naming-check).*?\))?:(?! Update CONTRIBUTORS.md)' --since="${LAST_EDIT_TIME}" --pretty="%s" > log_output.txt
 
       - name: Check if there are new commits
         run: |
--- a/.github/workflows/automatic-changelog.yml
+++ b/.github/workflows/automatic-changelog.yml
@@ -49,9 +49,10 @@ jobs:
         if: env.NEW_COMMITS == 'true'
         env:
           GITHUB_TOKEN: ${{ secrets.PAT }}
+          PAT: ${{ secrets.PAT }}
         run: |
           git config --global user.email "actions@github.com"
           git config --global user.name "GitHub Actions"
           git diff
           git commit -am "chore(changelog): Update updates.txt" || exit 0
-          git push https://x-access-token:${{ secrets.PAT }}@github.com/${{ github.repository }}.git HEAD:master
+          git push "https://x-access-token:${PAT}@github.com/${GITHUB_REPOSITORY}.git" HEAD:master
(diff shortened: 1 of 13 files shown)


rcdailey:
flagging @saltydk since you've done some security hardening in the past here.

.github/workflows/automatic-changelog.yml
```diff
@@ -48,9 +49,10 @@ jobs:
         if: env.NEW_COMMITS == 'true'
         env:
           GITHUB_TOKEN: ${{ secrets.PAT }}
```
FonduemangVI: `GITHUB_TOKEN` is now redundant. The `git push` below uses `${PAT}` explicitly, so this env var is unused unless a later git op relies on it. Could drop it and keep only `PAT`.

.pre-commit-config.yaml
```diff
@@ -89,3 +89,8 @@ repos:
     hooks:
       - id: editorconfig-checker
         name: EditorConfig Lint
+
+  - repo: https://github.com/zizmorcore/zizmor-pre-commit
+    rev: v1.25.0
```
FonduemangVI, line 94: This hook is pinned by tag (`rev: v1.25.0`) while the repo pins GitHub Actions by commit SHA. Tag-pinning is the pre-commit norm and fine, but it is inconsistent with the SHA-pinning discipline used everywhere else in this PR.

FonduemangVI (commented):
Reviewed the workflow hardening. Solid, well-scoped PR, template-injection fixes, least-privilege permissions, and the `persist-credentials: false` split are all correct. Two optional cleanups inline; nothing blocking.

.github/workflows/automatic-changelog.yml
rcdailey: Good catch, removed it.

.pre-commit-config.yaml
rcdailey: Tag pinning is the pre-commit convention and matches every other hook in this file. SHA pinning in pre-commit is unusual since hooks run locally in a sandboxed virtualenv rather than with repo secrets like Actions do. I think the current approach is fine, but happy to change if you feel strongly.

FonduemangVI (approved):
Good to go. Thanks for this

PR #2744, opened 2026-05-17
feat(starr): Change the default score for the CF `WiTH AD` to `-10k`

TRaSH-:
# Pull Request

## Purpose

After some research, we noticed that releases with Audio Description often have that track set as the default audio track. We had expected it to be added as a second audio track instead, allowing the user to choose.
In general, releases with Audio Description receive the same score, and if a release with Audio Description is released earlier, it will not upgrade to one without Audio Description. To prevent this, we decided to give this CF a default score of -10k for the larger user base.

## Approach

Changed the default score for the CF `WiTH AD` to `-10k`

## Requirements

## Summary by Sourcery

Enhancements:
- Lower the default score of the `WiTH AD` custom format in both Radarr and Sonarr JSON configs to reduce selection preference for audio-description releases.

### DIFF
--- a/docs/json/radarr/cf/with-ad.json
+++ b/docs/json/radarr/cf/with-ad.json
@@ -1,6 +1,11 @@
 {
   "trash_id": "127bdbadcf3e4463a8c707759fbaad75",
   "trash_regex": "https://regex101.com/r/YWHIIr/1",
+  "trash_scores": {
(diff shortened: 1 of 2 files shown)


quorn23 (approved):
Makes sense, as the people needing it will look for it specifically.

PR #2758, opened 2026-05-27
feat(sonarr): Add NEW Custom Format `BW` (Black & White)

TRaSH-:
# Pull Request

## Purpose

Some shows are released in both Color and Black & White versions. The default score for this Custom Format is set to -10000. Since Sonarr doesn’t support editions, and if you prefer the Black & White version, assign a positive score to this Custom Format. Sonarr will then prefer the Black & White releases, and the Custom Format will also be used in your naming scheme.

## Approach

- Added NEW Custom Format `BW`
- Added new Custom Format to the collection table
- Added new Custom Format to the unwanted groups

## Requirements

## Summary by Sourcery

Add documentation and configuration for a new Sonarr custom format representing Black & White releases.

New Features:
- Introduce a new Sonarr custom format `BW` for Black & White releases with associated JSON definition.

Enhancements:
- Add the new `BW` custom format to the Sonarr custom formats collection and index tables.
- Include the `BW` custom format in the unwanted formats custom format groups across supported locales.

Documentation:
- Document the new `BW` (Black & White) custom format, including description and usage guidance in the Sonarr custom formats guide.

### DIFF
--- a/docs/Sonarr/sonarr-collection-of-custom-formats.md
+++ b/docs/Sonarr/sonarr-collection-of-custom-formats.md
@@ -103,10 +103,11 @@ Special thanks to everyone who has helped in the creation and testing of these C
 | [Remaster](#remaster) | [Bad Dual Groups](#bad-dual-groups)     | [Remux Tier 02](#remux-tier-02)         |
 |                       | [BR-DISK](#br-disk)                     | [HD Bluray Tier 01](#hd-bluray-tier-01) |
 |                       | [BR-DISK (BTN)](#br-disk-btn)           | [HD Bluray Tier 02](#hd-bluray-tier-02) |
-|                       | [Extras](#extras)                       | [WEB Tier 01](#web-tier-01)             |
-|                       | [LQ](#lq)                               | [WEB Tier 02](#web-tier-02)             |
-|                       | [LQ (Release Title)](#lq-release-title) | [WEB Tier 03](#web-tier-03)             |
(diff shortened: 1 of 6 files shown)


headscarf737:
This is called `Black and White Editions` for radarr, why not have the same name?
this is the default for the unwanted german cf-group for radarr, why not make it the default for sonarr too?

TRaSH-:
> This is called `Black and White Editions` for radarr, why not have the same name? this is the default for the unwanted german cf-group for radarr, why not make it the default for sonarr too?

Because Sonarr officially doesn't support editions so we didn't go with Edition, and let's say you prefer the BW version of that show over the colored one, you also want the custom format to be used in your naming scheme. To stay in line with the scene's approach to these releases, we chose BW.

`Show S01 Authentic BW 2160p AMZN WEB-DL DDP 5.1 Atmos DV HDR H.265-RlsGrp`
`Show.S01.BW.DV.HDR.2160p.AMZN.WEB-DL.DDPA5.1.H.265-RlsGrp`
`Show S01 Authentic BW 1080p AMZN WEB-DL DDP 5.1 Atmos H.264-RlsGrp`
`Show.S01.BW.1080p.AMZN.WEB-DL.DDPA5.1.H.264-RlsGrp`

We didn't enable this CF by default in the unwanted group because some users might prefer the BW version. If they already have the BW version and we enable the BW CF by default, their preferred BW version could be scored with -10k during the next profile and custom format upgrade, and it might be replaced. To prevent this, we added it to the unwanted group but did not enable it by default. Users can do this manually to avoid unwanted upgrades or downgrades.

PR #2765, opened 2026-06-07
chore(gha): Skip changelog workflow on forks

rcdailey:
The automatic changelog workflow runs on a weekly schedule, but it fails on forks because `secrets.PAT` isn't available there. Forks don't need this workflow anyway since it auto-generates changelog entries and pushes directly to master.

Added a guard using `github.event.repository.fork` so the job is skipped on any fork automatically.


nuxencs (approved):
don't see anything wrong with that change

FonduemangVI (approved):

PR #2766, opened 2026-06-07
chore(build): Added labels for `Area: Guide` and `Area: Guide Sync` and cleaned up the backend label

TRaSH-:
# Pull Request

## Purpose

Added labels for `Area: Guide` and `Area: Guide Sync` and cleaned up the backend label

- Currently, the backend label also matches changes in the guide documents. With these updates, any changes to the written guide should be labeled `Area: Guide`, not `Area: Backend`.
- Any changes in the JSON folder are now marked with `Area: Guide Sync`.

## Approach

Added labels for `Area: Guide` and `Area: Guide Sync` and cleaned up the backend label

## Requirements

## Summary by Sourcery

Build:
- Update labeler rules to add Area: Guide and Area: Guide Sync labels and exclude guide docs and tooling from the backend label.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -37,19 +37,7 @@
           ["docs/json/radarr/cf/**", "docs/json/sonarr/cf/**"]
 "Area: Guide Sync":
   - changed-files:
-      - any-glob-to-any-file:
-          [
-            "docs/json/radarr/cf/**",
-            "docs/json/radarr/cf-groups/**",
-            "docs/json/radarr/naming/**",
-            "docs/json/radarr/quality-profiles/**",
-            "docs/json/radarr/quality-size/**",
-            "docs/json/sonarr/cf/**",
-            "docs/json/sonarr/cf-groups/**",
-            "docs/json/sonarr/naming/**",
-            "docs/json/sonarr/quality-profiles/**",
-            "docs/json/sonarr/quality-size/**",
-          ]
+      - any-glob-to-any-file: ["docs/json/**"]
 "Area: Starr Naming":
   - changed-files:
       - any-glob-to-any-file:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -26,7 +26,6 @@
   - changed-files:
       - any-glob-to-any-file:
           [
-            "docs/json/sonarr/rp/**",
             "docs/json/sonarr/cf/**",
             "docs/json/sonarr/quality-size/**",
             "docs/json/sonarr/naming/**",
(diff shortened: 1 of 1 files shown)


TRaSH-:
> * After removing `docs/json/sonarr/rp/**` from the backend label, it doesn’t appear to be covered by any new label; if those RP JSON files should still be auto-labeled, consider adding them under `Area: Guide Sync` or another appropriate section.

Sonarr release profiles are absolute, which is why I actually removed them. If we think we need it later, we can always add it back.

.github/labeler.yml
```diff
@@ -56,11 +70,30 @@
   - changed-files:
       - any-glob-to-any-file:
           [
+            ".vscode/**",
+            "overrides/**",
+            "scripts/**",
+            "!docs/assets/**",
+            "!docs/img/**",
+            "!docs/js/**",
+            "!docs/stylesheets/**",
```
rcdailey: These are prefixed with `!`, which makes them exclusion patterns. They're filtering out files that would otherwise match a positive pattern, but none of the positive patterns here (`.vscode/**`, `overrides/**`, `scripts/**`) match anything under `docs/` anyway.

Based on the discussion, it sounds like the intent was to *include* these paths (they're backend/infrastructure, not guide content). Drop the `!` prefix so they're positive matches:

```suggestion
            "docs/assets/**",
            "docs/img/**",
            "docs/js/**",
            "docs/stylesheets/**",
          ]
```

TRaSH-: Hmm, it seems I didn't understand what `!` means :(
It should actually match those folders, so let me fix it on my end.

.github/labeler.yml
```diff
@@ -36,6 +35,21 @@
...
+            "docs/json/radarr/cf-groups/**",
+            "docs/json/radarr/naming/**",
+            "docs/json/radarr/quality-profiles/**",
+            "docs/json/radarr/quality-size/**",
+            "docs/json/sonarr/cf/**",
+            "docs/json/sonarr/cf-groups/**",
+            "docs/json/sonarr/naming/**",
+            "docs/json/sonarr/quality-profiles/**",
+            "docs/json/sonarr/quality-size/**",
+          ]
```
rcdailey: This lists 10 specific subdirectories, but `docs/json/**` would match all of them plus the ones currently missed (`quality-profile-groups/`, `dummy-files/`, `conflicts.json`, and `docs/json/guide-only/`). It also wouldn't need updating when new subdirectories are added.

```suggestion
"Area: Guide Sync":
  - changed-files:
      - any-glob-to-any-file: ["docs/json/**"]
```

.github/labeler.yml
```diff
@@ -56,11 +70,29 @@
...
+            "docs/Glossary/**",
+            "docs/Guide-Sync/**",
+            "docs/Lidarr/**",
+            "docs/Misc/**",
+            "docs/Plex/**",
+            "docs/Prowlarr/**",
+            "docs/Radarr/**",
+            "docs/Recyclarr/**",
+            "docs/Sonarr/**",
           ]
```
rcdailey, line 86: Listing every directory explicitly means new ones (like `docs/SQP/`) and root-level files (`docs/index.md`, `docs/thnx.md`, `docs/user-promotion-deals.md`) get missed. It also needs updating whenever a new section is added.

`docs/**/*.md` would catch everything automatically. The only exclusion needed is `!docs/json/**` to avoid overlap with "Area: Guide Sync" (JSON is sync data, not written guide content). The section-specific labels like "Area: Radarr" can coexist with "Area: Guide" since those files are still guide content.

```suggestion
"Area: Guide":
  - changed-files:
      - any-glob-to-any-file:
          ["includes/**", "docs/**/*.md", "!docs/json/**"]
```

TRaSH-: This would mean only MD file changes are detected. Updated images or anything else in that folder aren't included. Additionally, there are folders inside that shouldn't get the guide label.

PR #2770, opened 2026-06-14
fix(downloaders): Normalize fclones version comparison

Pa7rickStar:
# Pull Request

## Purpose

Fixes an issue in the qBittorrent Cache Mover end script where the installed `fclones` version could be treated as outdated even when it matched the latest GitHub release.

The installed binary reports versions without a leading `v`, for example:

```text
0.35.0
```

GitHub release tags include the leading `v`, for example:

```text
v0.35.0
```

Because the script compared these values directly, it could reinstall `fclones` unnecessarily even when the installed version was already current.

## Approach

* Keep the GitHub release tag value for the download URL.
* Strip a leading `v` into a normalized version value.
* Compare the installed binary version against the normalized version.
* Continue using the original tag value for the GitHub release download path.

This avoids unnecessary reinstalls while preserving the correct download URL format.

None.

## Requirements

* [x] These changes meet the standards for [[contributing](https://github.com/TRaSH-Guides/Guides/blob/master/CONTRIBUTING.md)](https://github.com/TRaSH-Guides/Guides/blob/master/CONTRIBUTING.md).
* [x] I have read the [[code of conduct](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CODE_OF_CONDUCT.md)](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CODE_OF_CONDUCT.md).

## Summary by Sourcery

Normalize fclones version comparison in the qBittorrent Cache Mover end script to avoid unnecessary reinstalls and bump the script version metadata.

Bug Fixes:
- Ensure fclones version checks handle GitHub tags with a leading 'v' so the installed binary is not treated as outdated when it matches the latest release.

Enhancements:
- Update the qBittorrent Cache Mover end script version and metadata to reflect the latest changes.

### DIFF
--- a/includes/downloaders/mover-tuning-end.sh
+++ b/includes/downloaders/mover-tuning-end.sh
@@ -204,7 +204,7 @@ install_fclones_binary() {
     LATEST_VERSION=$($GITHUB_API_CMD 2>/dev/null | grep -Po '"tag_name": "\K.*?(?=")') || true
     if [[ -z "$LATEST_VERSION" ]]; then
         log "⚠ Could not fetch latest release, using default version $DEFAULT_VERSION (continuing anyway)"
-    LATEST_VERSION="v$DEFAULT_VERSION"
+        LATEST_VERSION="v$DEFAULT_VERSION"
     else
         log "Latest fclones release: $LATEST_VERSION"
     fi
--- a/includes/downloaders/mover-tuning-end.sh
+++ b/includes/downloaders/mover-tuning-end.sh
@@ -3,12 +3,12 @@ set -euo pipefail # Exit on error, undefined variables, and pipe failures
 
 # =====================================
 # Script: qBittorrent Cache Mover - End
-# Version: 1.3.1
-# Updated: 20260411
+# Version: 1.3.3
+# Updated: 20260614
 # =====================================
 
 # Script version and update check URLs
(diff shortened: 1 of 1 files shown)


TRaSH-:
@patrick-5546, could you review the sourcery-ai comments and respond or make any necessary fixes?

Pa7rickStar:
> * In `install_fclones_binary`, the line setting `LATEST_VERSION="v$DEFAULT_VERSION"` has been dedented out of the `if [[ -z "$LATEST_VERSION" ]]` block, which will now overwrite `LATEST_VERSION` with the default value even when the GitHub API call succeeds; it should remain indented inside that `if` so the default is only applied on failure.

Line 207 is wrongly indented indeed. This does however not lead to the described behavior. `LATEST_VERSION="v$DEFAULT_VERSION"` is still inside the `if [[ -z "$LATEST_VERSION" ]]` fallback block, before the `else`. Bash does not care about indentation. I fixed it anyways.

@TRaSH- should be good now.

TRaSH-:
@BZ00001 could you review this one?

BaukeZwart:
PR is looking solid, I will do a live test asap to check the comparison logic itself.

BaukeZwart:
I tested the PR, it's good to go.

TRaSH- (approved):
Approved by @BZ00001

PR #2787, opened 2026-06-21
chore(misc): Add 'Add Release Group' workflow_dispatch

bakerboy448:
## What

A **maintainer-run `workflow_dispatch`** (Actions → *Add Release Group* → Run workflow) that makes adding a single release group/title to a group-list Custom Format stupid-easy and opens the PR for you.

### Inputs
- **CF group** — dropdown of the common group-list CFs (bad-dual-groups, lq, the `*-lq`/`*-scene` lists, the LQ release-title lists) **+ a validated free-text override** for any other group-list CF (e.g. single-group or tier CFs).
- **group_name**, **reason** (required), optional **regex** override and **issue** number.

### Behaviour
- **Auto-detects** `ReleaseGroupSpecification` vs `ReleaseTitleSpecification` and the value shape (`^(name)$` for groups, `\b(name)\b` for titles) from the chosen CF — override available for special cases (e.g. `^(MGE\b.*)$`).
- Adds to **whichever apps** (Radarr/Sonarr) have that CF, inserts **alphabetically**, is **idempotent** (skips if already present), and **`check-jsonschema`-validates** before opening the PR.
- PR body includes the reason and `Closes #<issue>` when given.

Actions are SHA-pinned; documented in CONTRIBUTING under *Release Group … Additions*. Tested locally end-to-end (insert + alpha-sort + schema-valid + idempotent).

## Summary by Sourcery

Introduce a maintainer-run GitHub Actions workflow to add a single release group/title to Custom Format group-lists and automatically open a PR, and document its usage in CONTRIBUTING.

New Features:
- Add an `Add Release Group` workflow_dispatch GitHub Action that edits relevant Radarr/Sonarr CF group-list JSON, validates against schemas, and opens a pull request with prepared metadata.

Documentation:
- Document the Add Release Group workflow in CONTRIBUTING, including when and how to use it instead of manual JSON edits.

## AI disclosure

This pull request was created with AI assistance.

### DIFF
--- a/.github/workflows/add-release-group.yml
+++ b/.github/workflows/add-release-group.yml
@@ -53,6 +53,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
       - name: Setup Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
@@ -172,12 +174,14 @@ jobs:
 
       - name: Validate changed JSON against the CF schema
         if: steps.edit.outputs.did_change == 'true'
+        env:
+          CHANGED_FILES: ${{ steps.edit.outputs.files }}
         run: |
           while IFS= read -r f; do
             [ -z "$f" ] && continue
             app=$(printf '%s' "$f" | cut -d/ -f3)
             check-jsonschema --schemafile "schemas/${app}-cf.schema.json" "$f"
-          done <<< "${{ steps.edit.outputs.files }}"
+          done <<< "$CHANGED_FILES"
 
       - name: Open pull request
         if: steps.edit.outputs.did_change == 'true'
--- a/.github/workflows/add-release-group.yml
+++ b/.github/workflows/add-release-group.yml
@@ -1,0 +1,184 @@
+# Manually add a release group/title to a CF group-list and open a PR. Run from the Actions tab.
+name: Add Release Group
+
+on:
+  workflow_dispatch:
+    inputs:
+      cf_group:
+        description: "Which group-list Custom Format to add to"
+        required: true
+        type: choice
+        default: bad-dual-groups
+        options:
+          - bad-dual-groups
+          - lq
+          - lq-release-title
+          - anime-lq-groups
+          - french-lq
+          - french-scene
+          - german-lq
+          - german-scene
+          - german-anime-scene
+          - german-microsized
+          - web-scene
+          - other
+      cf_other:
+        description: "If 'other' above: the CF json basename (e.g. flux). Validated to be a group-list CF."
+        required: false
+        type: string
+      group_name:
+        description: "Release group / title to add (e.g. C76)"
+        required: true
+        type: string
+      regex:
+        description: "Optional regex override. Blank = auto: ^(name)$ for groups, \\b(name)\\b for titles."
+        required: false
+        type: string
+      reason:
+        description: "Reason / description (goes in the PR body) — required"
+        required: true
+        type: string
+      issue:
+        description: "Issue number to reference, e.g. 2777 (optional)"
+        required: false
+        type: string
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  add-release-group:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Setup Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.x"
+
+      - name: Install validator
+        run: pip install check-jsonschema
+
+      - name: Add the group and prepare PR metadata
+        id: edit
+        env:
+          CF_GROUP: ${{ inputs.cf_group }}
+          CF_OTHER: ${{ inputs.cf_other }}
+          GROUP_NAME: ${{ inputs.group_name }}
+          REGEX: ${{ inputs.regex }}
+          REASON: ${{ inputs.reason }}
+          ISSUE: ${{ inputs.issue }}
+        run: |
+          python3 - <<'PY'
+          import json, os, re, glob
+
+          cf = os.environ["CF_GROUP"].strip()
+          if cf == "other":
+              cf = os.environ["CF_OTHER"].strip()
+          name = os.environ["GROUP_NAME"].strip()
+          override = os.environ.get("REGEX", "").strip()
+          reason = os.environ.get("REASON", "").strip()
+          issue = os.environ.get("ISSUE", "").strip().lstrip("#")
+
+          def die(msg):
+              print(f"::error::{msg}")
+              raise SystemExit(1)
+
+          if not cf:
+              die("No CF group selected — pick one or set the 'other' basename.")
+          if not name:
+              die("group_name is required.")
+          if not reason:
+              die("reason/description is required.")
+
+          apps = [a for a in ("radarr", "sonarr") if os.path.isfile(f"docs/json/{a}/cf/{cf}.json")]
+          if not apps:
+              valid = sorted({os.path.basename(p)[:-5] for p in glob.glob("docs/json/*/cf/*.json")})
+              die(f"CF '{cf}' not found. Existing CF basenames: {', '.join(valid)}")
+
+          GROUP_IMPLS = {"ReleaseGroupSpecification", "ReleaseTitleSpecification"}
+          changed, impl_used = [], None
+          for a in apps:
+              path = f"docs/json/{a}/cf/{cf}.json"
+              with open(path, encoding="utf-8") as fh:
+                  data = json.load(fh)
+              specs = data.get("specifications", [])
+              impls = {s.get("implementation") for s in specs}
+              if impls and not impls <= GROUP_IMPLS:
+                  die(f"{cf} is not a release-group/title list (implementations: {sorted(impls)}).")
+              impl = next(iter(impls), "ReleaseGroupSpecification")
+              impl_used = impl
+              if any(s.get("name", "").lower() == name.lower() for s in specs):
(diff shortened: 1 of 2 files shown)


TRaSH-:
@bakerboy448 Could you please review the comment from Sourcery-AI?

bakerboy448:
Addressed in `6438bfb5` — applied Sourcery's suggestion. The workflow now records the first changed app's implementation and **fails with a clear error** if another changed app uses a different one (`ReleaseGroupSpecification` vs `ReleaseTitleSpecification`), instead of silently letting the last-processed app dictate `kind`/title/body. So mixed-implementation setups are now deterministic (they error and ask you to align) rather than misleading.

.github/workflows/add-release-group.yml
```diff
@@ -0,0 +1,184 @@
...
+          for a in apps:
+              path = f"docs/json/{a}/cf/{cf}.json"
+              with open(path, encoding="utf-8") as fh:
+                  data = json.load(fh)
+              specs = data.get("specifications", [])
+              impls = {s.get("implementation") for s in specs}
+              if impls and not impls <= GROUP_IMPLS:
+                  die(f"{cf} is not a release-group/title list (implementations: {sorted(impls)}).")
+              impl = next(iter(impls), "ReleaseGroupSpecification")
+              impl_used = impl
```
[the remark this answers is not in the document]
bakerboy448: Addressed in `6438bfb5`. The workflow now records the first changed app's implementation and **fails with a clear error** if another changed app uses a different one (`ReleaseGroupSpecification` vs `ReleaseTitleSpecification`), so `kind`/title/body are deterministic instead of last-app-wins. Thanks for the catch!

.github/workflows/add-release-group.yml
```diff
@@ -0,0 +1,194 @@
...
+
+          apps = [a for a in ("radarr", "sonarr") if os.path.isfile(f"docs/json/{a}/cf/{cf}.json")]
+          if not apps:
+              valid = sorted({os.path.basename(p)[:-5] for p in glob.glob("docs/json/*/cf/*.json")})
+              die(f"CF '{cf}' not found. Existing CF basenames: {', '.join(valid)}")
+
+          GROUP_IMPLS = {"ReleaseGroupSpecification", "ReleaseTitleSpecification"}
+          changed, impl_used = [], None
+          for a in apps:
+              path = f"docs/json/{a}/cf/{cf}.json"
```
FonduemangVI, line 117: **Path traversal via unsanitized `cf_other` can write outside `docs/json/*/cf/`.**

When `cf_group == "other"`, `cf` is taken verbatim from the free-text `cf_other` input (line 82) with only `.strip()` applied. It then flows straight into a filesystem path:

```python
apps = [a for a in ("radarr", "sonarr") if os.path.isfile(f"docs/json/{a}/cf/{cf}.json")]
...
path = f"docs/json/{a}/cf/{cf}.json"
with open(path, encoding="utf-8") as fh:
    data = json.load(fh)
...
with open(path, "w", encoding="utf-8") as fh:
    json.dump(data, fh, indent=2, ensure_ascii=False)
```

Nothing stops `cf_other` containing `../`. If someone passes e.g. `../../../schemas/radarr-cf` (or any path that, after traversal, points at an existing JSON file matching the implementation check), `os.path.isfile` will happily find it, `json.load` will parse it, and the script will then `json.dump` a mutated copy back over that file outside the intended `docs/json/<app>/cf/` directory entirely.

The input description says `cf_other` is "Validated to be a group-list CF," but the actual validation is only: file-exists-at-that-path (line 99) + implementation-membership (line 112). There's no basename/traversal check.

This requires `workflow_dispatch` access (write-permission users only), so it's not a remote/unauthenticated vector but it's still an easy way for a typo or a copy-pasted path to corrupt an unrelated file in the repo, and worth closing off given the field is explicitly free text. Suggested fix:

```python
cf = os.environ["CF_OTHER"].strip()
if cf != os.path.basename(cf) or not re.fullmatch(r"[A-Za-z0-9._-]+", cf):
    die(f"Invalid CF name '{cf}' must be a bare basename, no path separators.")
```

.github/workflows/add-release-group.yml
```diff
@@ -0,0 +1,194 @@
...
+              with open(path, encoding="utf-8") as fh:
+                  data = json.load(fh)
+              specs = data.get("specifications", [])
+              impls = {s.get("implementation") for s in specs}
+              if impls and not impls <= GROUP_IMPLS:
+                  die(f"{cf} is not a release-group/title list (implementations: {sorted(impls)}).")
+              impl = next(iter(impls), "ReleaseGroupSpecification")
```
FonduemangVI: **Non-deterministic implementation choice when a CF's specs mix both `ReleaseGroupSpecification` and `ReleaseTitleSpecification`.**

```python
impls = {s.get("implementation") for s in specs}
if impls and not impls <= GROUP_IMPLS:
    die(...)
impl = next(iter(impls), "ReleaseGroupSpecification")
```

`impls` is a Python `set`. The validation on line 112 explicitly *permits* a file containing both implementations (since `{"ReleaseGroupSpecification", "ReleaseTitleSpecification"} <= GROUP_IMPLS` is true). In that case `next(iter(impls))` picks whichever of the two the set happens to iterate first and since Python randomizes string hash seeds per process (`PYTHONHASHSEED` unset by default), that order isn't stable across workflow runs. The regex shape chosen for the new entry (`^(name)$` vs `\b(name)\b`) would then depend on hash-seed luck rather than the file's actual content.

Checked the current dropdown CFs (`bad-dual-groups`, `lq`, `lq-release-title`, `anime-lq-groups`) against the repo right now none of them currently mix implementations, so this doesn't bite today. But `cf_other` opens this up to any group-list CF, including future ones, so it's a live foot-gun rather than a theoretical one. Suggest either rejecting mixed-implementation files outright, or picking deterministically (e.g. majority vote, or first spec in file order instead of a set).

.github/workflows/add-release-group.yml
```diff
@@ -0,0 +1,194 @@
...
+              reason,
+          ]
+          if issue:
+              body_lines += ["", f"Closes #{issue}"]
+          body = "\n".join(body_lines)
+
+          for key, val in (("did_change", "true"), ("branch", branch), ("title", title)):
+              out.write(f"{key}={val}\n")
+          out.write("body<<BODYEOF\n" + body + "\nBODYEOF\n")
+          out.write("files<<FILESEOF\n" + "\n".join(p for _, p in changed) + "\nFILESEOF\n")
```
FonduemangVI: **`GITHUB_OUTPUT` multiline write uses a static delimiter collides if `reason` (or `group_name`, via `title`/`branch`) contains a matching line.**

```python
out.write("body<<BODYEOF\n" + body + "\nBODYEOF\n")
out.write("files<<FILESEOF\n" + "\n".join(p for _, p in changed) + "\nFILESEOF\n")
```

`body` embeds the free-text `reason` input verbatim (line 161). If a maintainer's `reason` text happens to contain a line that's exactly `BODYEOF`, the heredoc-style `GITHUB_OUTPUT` write terminates early at that line, and everything after it in `reason` gets parsed by the Actions runner as additional `key=value` output lines instead of being part of `body` output injection into the step's outputs (which downstream steps consume for the PR title/body/branch).

This is exactly the failure mode GitHub's own docs warn about for multiline outputs, which is why the recommended pattern is a randomly generated delimiter (e.g. `delimiter=$(uuidgen)` or similar) instead of a fixed string. Low probability someone types `BODYEOF` literally in a reason field, but it's a one-line fix to make it not depend on that:

```python
import uuid
delim = f"EOF_{uuid.uuid4().hex}"
out.write(f"body<<{delim}\n" + body + f"\n{delim}\n")
```

(same for `files`, though that one's built from repo-controlled paths so lower risk).

.github/workflows/add-release-group.yml
FonduemangVI: **Separate, smaller edge case on the same line: default-implementation guess is wrong for a CF with an empty `specifications` list.**

`next(iter(impls), "ReleaseGroupSpecification")` defaults to group-type whenever `impls` is empty i.e. whenever the target CF file currently has zero specifications. That's plausible for a brand-new, still-empty group-list CF reached via `cf_other`. If that CF is actually meant to hold titles (`ReleaseTitleSpecification`), the very first entry added through this workflow gets the group regex shape `^(name)$` instead of the title shape `\b(name)\b` silently wrong, and nothing downstream (schema validation) would catch a mismatch between implementation type and how the regex is anchored, since the schema doesn't cross-check that.

Minor compared to the other three, flagging since it's adjacent to the mixed-implementation issue above and shares the same root cause (guessing implementation type from `specifications` content instead of requiring/recording it explicitly).

FonduemangVI (commented):
Ran a review pass on this (correctness + security angles, focused on the new `add-release-group.yml` script since it's the whole surface area of the PR). Left 4 inline comments one likely worth fixing before merge (path traversal on `cf_other`), three lower-probability edge cases worth a quick look. Nice touch overall: SHA-pinned actions, `persist-credentials: false`, schema validation before opening the PR, and env-var passing into the Python heredoc (avoids shell-interpolation injection in the common case).

.github/workflows/add-release-group.yml
bakerboy448: Fixed in `4c2463701`. `cf_other` now must match the repository basename convention before any filesystem lookup, so path separators and traversal components are rejected.

.github/workflows/add-release-group.yml
bakerboy448: Fixed in `4c2463701`. A target file must contain exactly one supported implementation. Mixed release-group/title specifications now fail explicitly instead of depending on set iteration order.

.github/workflows/add-release-group.yml
bakerboy448: Fixed in `4c2463701`. Multiline outputs now use a UUID-based delimiter with an explicit collision check. `group_name` is also restricted to one line before it reaches the single-line title output.

.github/workflows/add-release-group.yml
bakerboy448: Fixed in `4c2463701`. Empty specification lists now fail with a clear error because the implementation cannot be inferred safely.

FonduemangVI (approved):
Thanks for this. Nice improvement that should remove some friction from people helping out

PR #2789, opened 2026-06-21
fix(ci): prevent fork-PR code execution in privileged deploy workflow

rcdailey:
### Purpose

The PR deploy workflow chain (`trigger-pr-deploy.yml` + `deploy-pr.yml`) checks out fork-PR code and runs `pip install` in the privileged `workflow_run` job that has access to `CF_API_TOKEN` and `GITHUB_TOKEN`. A malicious fork could modify `requirements.txt` to execute arbitrary code during `pip install` and exfiltrate those secrets.

This is the same vulnerability class as [CVE-2026-55675](https://github.com/saltyorg/docs/security/advisories/GHSA-wr2h-fpxm-ff4x) (CVSS 10.0), which was found and fixed in `saltyorg/docs`. Credit to @saltydk for flagging it here.

### Approach

Moved the entire build (checkout, `pip install`, `mkdocs build`) into the unprivileged `trigger-pr-deploy.yml` workflow, which runs on `pull_request` with no secret access. The built `site/` directory gets bundled with PR metadata into a single artifact (`pr_build`).

`deploy-pr.yml` is reduced to downloading that pre-built artifact and deploying the static HTML to Cloudflare. No fork code executes in the secret-bearing job.

Specifically:

- `trigger-pr-deploy.yml` now performs: path filtering, checkout (sparse), setup-python, pip install, mkdocs build, then uploads the built site + PR metadata as a single artifact
- `deploy-pr.yml` loses its `check` and `build` jobs; the single `deploy` job downloads the artifact, reads metadata, and deploys to Cloudflare (gated on `workflow_run.conclusion == 'success'`)
- The `comment` job still runs on `always()` so it can post failure comments, and no longer checks out the repo (it only runs curl)
- All existing zizmor hardening is preserved (permissions blocks, env vars, suppression comments)

### Requirements

## Summary by Sourcery

Move PR preview build into an unprivileged workflow and limit the privileged deploy workflow to deploying pre-built artifacts to avoid executing fork code with secrets.

CI:
- Consolidate the deploy workflow into a single job that reads PR metadata from a pre-built artifact and deploys to Cloudflare only after successful builds.
- Shift checkout, dependency installation, and mkdocs build steps into the trigger workflow, bundling the built site and PR metadata into a single artifact consumed by the deploy workflow.
- Update PR comment logic to reflect deploy status and to rely on outputs from the new deploy job.

### DIFF
--- a/.github/workflows/deploy-pr.yml
+++ b/.github/workflows/deploy-pr.yml
@@ -1,4 +1,4 @@
-name: Build PR & Deploy
+name: Deploy PR Preview
 
 # workflow_run is intentional; this workflow deploys PR previews triggered by a
 # separate unprivileged workflow to safely access secrets.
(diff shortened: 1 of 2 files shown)


saltydk:
LGTM

FonduemangVI (approved):
Thanks for this one

PR #2848, opened 2026-08-08
fix(radarr): CF `LQ (Release Title)` to avoid false positives when matching in file and folder names.

TRaSH-:
# Pull Request

## Purpose

Fix Custom Format `LQ (Release Title)` to avoid false positives when matching in file and folder names.

The used regex pattern was unreliable because it only checks that a year appears somewhere before the match, not that the match itself is the release group.

The new regex pattern matches directly where a release group name actually sits, which is at the very end of the string, right after the final hyphen.

## Approach

Fix Custom Format `LQ (Release Title)` to avoid false positives when matching in file and folder names.

## Requirements

## Summary by Sourcery

Update Radarr custom format configuration for low-quality release titles to improve accuracy of release group detection.

Bug Fixes:
- Adjust the `LQ (Release Title)` custom format pattern to prevent false positives when matching file and folder names in Radarr.

Documentation:
- Update the Radarr custom format JSON definition for `LQ (Release Title)` to reflect the corrected matching pattern.

### DIFF
--- a/docs/json/radarr/cf/lq-release-title.json
+++ b/docs/json/radarr/cf/lq-release-title.json
@@ -42,7 +42,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "(?<=\\b[12]\\d{3}\\b.*?)(?<!\\b(web[ ._-]?(dl|rip)?).*?)\\b(EVO)\\b"
+        "value": "(?<!\\bweb[ ._-]?(dl|rip)?\\b.*)-(EVO)\\b"
       }
     },
     {
@@ -87,7 +87,7 @@
       "negate": false,
       "required": false,
       "fields": {
-        "value": "(?<=\\b[12]\\d{3}\\b.*?)(?<!\\b(web[ ._-]?(dl|rip)?).*?)\\b(PiRaTeS)\\b"
+        "value": "(?<!\\bweb[ ._-]?(dl|rip)?\\b.*)-(PiRaTeS)\\b"
       }
     },
     {


rg9400:
I'm not able to grasp the purpose of this change. If we are trying to *only* match after the hyphen, then why are we putting it in the release title condition? Sonarr and Radarr will be able to parse it as a release group, and we are much better served by using the release group condition and then the web quality/source. 

On the other hand, if we are saying that there are too many false positives, we can look at those filenames and see how to modify the existing regex a bit. Probably put a $ to anchor it at the end of the line instead.

nuxencs:
> I'm not able to grasp the purpose of this change. If we are trying to only match after the hyphen, then why are we putting it in the release title condition? Sonarr and Radarr will be able to parse it as a release group, and we are much better served by using the release group condition and then the web quality/source.

That would need to be a separate CF, because we can't combine different specifications as `OR`.

The issue is that we need to combine this into one single condition for each group and quality/source combination to keep it inside of a single CF.

> On the other hand, if we are saying that there are too many false positives, we can look at those filenames and see how to modify the existing regex a bit. Probably put a $ to anchor it at the end of the line instead.

I agree with that as well, we need more test cases imo.

TRaSH-:
Currently, a rename could trigger a false positive because of the annoying bug when the full path is shown.
image

But after this fix this won't happen for this specific use case
image

And the regex test case is also updated to show that this works on the single files
https://regex101.com/r/8TIqc2/latest

> .... Probably put a $ to anchor it at the end of the line instead.

Then it won't match when the release shows RlsGrp.mkv

TRaSH-:
Merging this one. If we want to review others, we can look at them later.

PR #1578, opened 2023-10-07
fix(starr): BR-DISK CF regex

nuxencs:
# Pull Request

## Purpose

Fix `BR-DISK` CF to match the edge case: `[BD]Movie.Title.2023.1080p.COMPLETE.BLURAY-RlsGrp`

## Approach

Adjust the `BR-DISK` regex and bring it up-to-date with the Radarr one.

## Requirements

- [x] These changes meet the standards for [contributing](https://github.com/TRaSH-Guides/Guides/blob/master/CONTRIBUTING.md).
- [x] I have read the [code of conduct](https://github.com/TRaSH-Guides/Guides/blob/master/.github/CODE_OF_CONDUCT.md).

PR #2878, opened 2026-09-03
fix(starr): BR-DISK should match 720p full-disc releases

jameskoziol:
# Pull Request

## Purpose

`BR-DISK` (Sonarr and Radarr) and `BR-DISK (BTN)` never match a full-disc release whose title carries `720p`, because `720p` sits in the regex's leading negative lookahead. Blu-ray discs can be authored at 720p, and at least one tracker names disc sets that way, for example `Series S01 720p Blu-ray AVC DTS-HD MA 2.0-RlsGrp` (a 288 GB BDMV folder set). Sonarr parses that as `Bluray-720p`, the CF stays silent, and with a positive audio score the pack is grabbed and then fails to import.

## Approach

Remove the `720p` alternative from the negative lookahead in all three definitions. Nothing else changes:

- 720p encodes remain unmatched through the existing `x264`/`x265`, `BDRip`, `MKV` and `REMUX` exemptions.
- The change is consistent with how the CF already treats `1080p Blu-ray AVC` titles as discs.
- All 23 existing regex101 test cases keep their result; sixteen 720p cases were added, anonymised per the guidelines (ten disc-set patterns across AVC, VC-1 and MPEG-2 that now match, plus x264, REMUX, BDRip and WEB-DL titles that must not).

regex101 (.NET flavour, all cases): https://regex101.com/r/NnBZx0/1

- [ ] If `720p` was added in #1578 for a false-positive case I could not find, I am happy to narrow this to a lookahead that only exempts `720p` when an encode marker is also present.

## Requirements

## AI disclosure

This pull request was created with AI assistance.

## Summary by Sourcery

Bug Fixes:
- Allow the Radarr and Sonarr BR-DISK custom formats to match full-disc Blu-ray releases labeled 720p without changing matching behavior for encoded or remuxed releases.

### DIFF
--- a/docs/json/radarr/cf/br-disk.json
+++ b/docs/json/radarr/cf/br-disk.json
@@ -5,7 +5,7 @@
     "german": -35000,
     "german-anime": -35000
   },
-  "trash_regex": "https://regex101.com/r/m6tJzR/1",
+  "trash_regex": "https://regex101.com/r/NnBZx0/1",
   "name": "BR-DISK",
   "includeCustomFormatWhenRenaming": false,
   "specifications": [
--- a/docs/json/radarr/cf/br-disk.json
+++ b/docs/json/radarr/cf/br-disk.json
@@ -5,7 +5,7 @@
     "german": -35000,
     "german-anime": -35000
   },
-  "trash_regex": "https://regex101.com/r/UpA3I7/2",
+  "trash_regex": "https://regex101.com/r/m6tJzR/1",
   "name": "BR-DISK",
   "includeCustomFormatWhenRenaming": false,
   "specifications": [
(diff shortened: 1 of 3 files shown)


TRaSH-:
is this a BTN specific thing? 

720p video is technically supported by the Blu-ray Disc specification, but commercial movie discs are almost never released in native 720p. Main feature films on official Blu-ray are mastered in 1080p or 4K.

Where 720p Appears on Blu-ray

- Bonus Features: Special features, making-of featurettes, or trailers included on a Blu-ray disc are sometimes encoded in 720p or standard definition (480p) depending on how they were originally filmed or archived.
- 3D Content: Some niche 3D Blu-ray formats or older 3D specifications utilized 720p at higher frame rates (like 60fps per eye) because full 1080p 3D was limited to 24fps.
- User-Created Discs: If you burn your own home videos onto a recordable Blu-ray disc (BD-R) using authoring software, you can choose to encode and output the video files in 720p.

jameskoziol:
Not BTN-specific, no. It is specific to a class of disc: content that was *produced* at 720p and then pressed to Blu-ray. Your list of where 720p shows up on disc is right, and I'd add the two biggest sources: US sports networks (several broadcast 720p60, and their disc releases are mastered from that), and concert / documentary / low-budget features shot in 720p. Studio feature films, agreed, almost never. But those disc classes are exactly what BR-DISK exists to stop, and they are the ones it currently misses.

**What actually happened to me**

Sonarr, TRaSH `WEB-2160p` profile with BR-DISK synced at -10000. I added a sports-documentary series and the search-on-add grabbed:

`Series S01 720p Blu-ray AVC DTS-HD MA 2.0-RlsGrp` — 287.9 GB

- Sonarr parsed it as `Bluray-720p`; BR-DISK bailed on `720p` before the `Blu-ray … AVC` check ran; DTS-HD MA gave it +2500, so it won.
- 30 queue entries, every one `00000.m2ts: Invalid season or episode`, 269 GiB on disk.
- I checked the files rather than trust the title: six BD-50s with `AACS`/`ANY!` folders (a pressed retail set from 2011, not a BD-R), and `ffprobe` on the feature streams gives `h264, 1280x720, 59.94 fps, progressive`. Native 720p60 straight off the broadcast master. The disc database lists this set as 1080p, which is wrong.

The same set also circulates without the group tag, and BR-DISK (BTN) misses it for the same reason.

**Other titles in the same class, from a quick search across my indexers** (names anonymised per the guidelines; sizes are the full BDMV set)

| Title | Size |
|---|---|
| `Series S02 720p USA Blu-ray AVC DTS-HD MA 2.0-RlsGrp` | 471 GB |
| `Movie 2016 720p Blu-ray AVC DTS-HD MA 2.0-RlsGrp` | 367 GB |
| `Series S01-02 720p Blu-ray VC-1 DTS-HD MA 5.1-RlsGrp` | 90 GB |
| `Series AKA Series 1987 S01 720p Blu-Ray AVC DTS-HD MA 2.0-RlsGrp` | 47 GB |
| `Movie 2014 720p Blu-ray MPEG-2 LPCM 2.0-RlsGrp` | 48 GB |
| `Movie 2010 720p Blu-ray AVC DTS-HD MA 5.1` | 24.6 GB |
| `Movie 2006 720p USA Blu-ray MPEG-2 DD 2.0-RlsGrp` | 33 GB |
| `Movie 2015 720p USA Blu-ray MPEG-2 DD 2.0-RlsGrp` | 21 GB |

Plus a run of 720p MPEG-2 concert discs and a few indie horror titles. None of these match BR-DISK today; all of them match with `720p` out of the lookahead.

**On scope**

I looked for a 720p title that this change would newly flag *wrongly* and couldn't construct one: every 720p encode carries x264/x265/H.264, BDRip, MKV or REMUX, and all of those exemptions are untouched (the regex101 link has the cases). The `720p` entry only changes the outcome for titles that say Blu-ray plus AVC/VC-1/MPEG-2 with no encoder tag, which is the disc signature, and it is already how the CF treats the identical 1080p title. If you'd rather keep `720p` in the list and add an explicit disc branch for 720p instead, I'm happy to rework it that way; the result is the same set of matches, just stated more defensively.

*(Edited: titles anonymised per the contributing guidelines and added to the regex101 cases.)*

TRaSH-:
if you could add them according to the guide guidelines and also change the ones in your above message. 
we normally don't discuss this kind of things on GitHub got a good reason

jameskoziol:
Done, and apologies - I had missed the guideline. I had anonymized some of the evidence data but clearly not enough.

TRaSH- (approved):
I performed several interactive search tests and didn't encounter any unexpected outcomes.
Since this is difficult to test and requires real-world environment testing, I approved it. We can always revert if we receive error reports or issues.
Thank you for the report and PR.

